Issue #19, opened 2017-11-14
Demoable Ego


andricDu:
The goal of this is to have a publicly available instance of Ego with which any user can log into an play with. 

## Tasks

- [x] Allow users to log in and be assigned admin rights
- [x] Do not persist users logging in to Ego into the Users table (to preserve privacy of users interacting with the demo)
- [x] Make behavior mentioned above toggle-able from a config value inside the `application.yml`
- [x] Script to auto-populate database with dummy users
- [x] Cronjob that refreshes the user database nightly to clean database of changes from users

## Resources
[Ego Source Code](https://github.com/overture-stack/ego)
[Ego UI Source Code](https://github.com/kids-first/kf-drp-auth-ui)

andricDu:
Let me know if you need any resources. At some point you might want an Amazon RDS instance for a demo postgres database.

joneubank:
Above commit ads a Spring Profile to enable the demo features. Run the application with the demo profile active.

Add the following command line option when running the jar:
`-spring.profiles.active=prod`

Note: activating any profile removes the "default" profile, which will cause an error in our case. Ensure you also run the "default" or "auth" profile to avoid issues.

joneubank:
Demo server running with nginx, https configured, and running the ego api locally with the demo and auth profile active.

https://demo.ego.kfdrc.org/swagger-ui.html

joneubank:
Cronjob added to execute the user dummy-data script at midnight daily. 200 randomly generated users will be available in the DB.

Next Step (as discussed with @cheapsteak ): Create dummy-data for applications and groups, including user relations to each.

joneubank:
Had to also make users "Approved" status before registration would work. Demo registered users will be approved by default and can now access the API's.

joneubank:
A lot of this work got overwritten by the JPA refactoring. I am redoing it now.

PR #20, opened 2017-11-14
[EGO-19] Demo Config and Dummy User Data

joneubank:
- Created "demo" profile, when active this will:
  - Allow users to log in and be assigned admin rights
  - Not persist users in the Users table (to preserve privacy of users interacting with the demo)
- Added script to populate database with dummy users

After merging, will need to configure cron job to run the sql script to reset the user table.

See note in https://github.com/overture-stack/ego/issues/19 about running the application with the demo profile active.


src/main/java/org/overture/ego/controller/AuthController.java
```diff
@@ -50,12 +49,18 @@
   @Autowired
   FacebookTokenService facebookTokenService;
 
+  // -- DEMO PROFILE CONFIGURATIONS --
+  @Value("${demo.user.save:true}")
```
ra-ver: Could we have just one property called demo

demo will be either true or false

src/main/java/org/overture/ego/controller/AuthController.java
```diff
@@ -92,15 +97,19 @@ boolean verifyJWToken(
...
+        // Saving user controlled through config - provides mechanism to run demos without saving
+        userService.create(user);
+      }
     }
     return tokenService.generateToken(user);
   }
 
   private User createNewUser(String userName, Map authInfo) {
-    String role = "USER";
+    String role = newUserRole;
```
ra-ver: could we do this rather.. demo == true ? "ADMIN" : "USER"

Also; could you please create constant for these Strings. Sorry it wasn't done in previous versions.

src/main/java/org/overture/ego/controller/AuthController.java
```diff
@@ -92,15 +97,19 @@ boolean verifyJWToken(
   private String generateUserToken(Map authInfo){
     val userName = authInfo.get("email").toString();
     User user = userService.getByName(userName, false);
+
     if (user == null) {
       user = createNewUser(userName,authInfo);
-      userService.create(user);
+      if (saveUsersAllowed) {
```
ra-ver: We'll need to do it for facebooktokens as well

joneubank: Controlling user creation at this point should work for facebook and google - This is in the `generateUserToken(...)` function called by both `exchangeFacebookTokenForAuth(...)` and `exchangeGoogleTokenForAuth(...)` .

joneubank:
@ra-ver Thanks. Comments addressed.

cheapsteak:
So psyched for this!
Could we please also add some dummy apps and groups as well?

andricDu:
https://github.com/overture-stack/ego/issues/19

PR #33, opened 2017-11-16
[EGO-19] Demo User registrations now Approved by default

joneubank:
API requests are rejected by secure filter if the user is in Pending state.

Updated SecureAuthorizationManager to use constants instead of String literals, and used the "demo" config value to assign Approved status to new users.

https://github.com/overture-stack/ego/issues/19


src/main/java/org/overture/ego/security/SecureAuthorizationManager.java
```diff
@@ -55,6 +55,6 @@ public boolean authorizeWithApplication(@NonNull Authentication authentication,
   }
 
   public boolean isActiveUser(User user){
-    return "approved".equals(user.getStatus().toLowerCase());
+    return UserStatus.APPROVED.toString().equals(user.getStatus());
```
andricDu: 👍

PR #34, opened 2017-11-22
PageDTO Added - Spring Page replaced in all Controllers

joneubank:
To keep with the existing API interface that was defined before inclusion of Spring Pagination. PageDTO provides the specific interface, and is built directly from the Spring Page results pro
vided from the services/repositories.


ra-ver:
Looks good.

Issue #31, opened 2017-11-16
Continuous Deployment for Ego


joneubank:
- [x] Enable Ego project in Circle CI
- [x] Add .circleci/config.yml
- [x] Add deployment config step for develop branch to demo server
- [x] Add deployment script configurable via environment variables
- [x] Add environment variable script template
- [x] Configure demo server to execute environment variable script

joneubank:
Added `src/main/resources/scripts` directory with scripts to start and stop the server using the java wrapper bin file. The start script takes environment variables to configure the application.

All required environment variables for the start and stop script are listed in `evn_template.sh`, stored also in the scripts directory . To configure your local or a server environment:
- Make a copy called `env.sh` (Note that this filename is added to the .gitignore)
- Populate all values needed in the copy
- Add the path to this file on the deployment server into the CircleCI environment variables
- Optionally, modify the server user's `~/.profile` file to have `source /path/to/env.sh` so that the variables load by default.

The `.circleci/deploy.sh` script includes a list of steps needed to take before deployment will work successfully. This includes setting up environment variables in Circle CI and on the deploy server, and creating the install paths on the deploy server.

PR #37, opened 2017-11-23
[Ego-31] Skip Maven tests during deploy job

joneubank:
Maven failed tests during package step of deploy. Requiring it to skip tests will prevent this in the future. Tests are done separately so are not required in the deploy step.

https://github.com/overture-stack/ego/issues/31


andricDu:
LGTM 👍

Issue #30, opened 2017-11-16
Dummy Data for Groups and Apps


joneubank:

joneubank:
Added a bash script that will execute all dummy-data files in sequence. Running this will clear the DB and set dummy-data in all tables. Example usage is included at top of the file:

./reset-dummy-data.sh --path='/srv/ego/dummy-data' --user='dbuser' --log='/path/to/logfile.log'

PR #39, opened 2017-11-23
[Ego-30] Dummy Data

joneubank:
SQL files provided for each Model and Relation.
Files will remove all current content and replace with dummy values.
Also includes a bash script to execute all SQL files in sequence - usable for cron-job on demo server.

Written for PSQL, won't work in H2.

https://github.com/overture-stack/ego/issues/30


joneubank:
Commits from another project got in here and need revision.

PR #45, opened 2017-11-24
Update Last-Login time when user token generated

joneubank:


joneubank:
WAIT TO MERGE

Will make this asynchronous - do not want returning JWT to wait on a SQL write.

PR #51, opened 2017-11-27
DB Host and Port added to env template

joneubank:
Also updated start script to use the new env vars.


joneubank:
@cheapsteak already have updated this server ;)

PR #52, opened 2017-11-27
Added support for filtering results by any field of the entity

ra-ver:
Added support for filtering by entity fields. User can apply filters to any Endpoint that returns a Page . Filters are applied as query parameters in the format: fieldName=fieldValue

### DIFF
--- a/src/main/java/org/overture/ego/controller/resolver/FilterResolver.java
+++ b/src/main/java/org/overture/ego/controller/resolver/FilterResolver.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2017. The Ontario Institute for Cancer Research. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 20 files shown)


src/main/java/org/overture/ego/controller/resolver/FilterResolver.java
```diff
@@ -0,0 +1,65 @@
...
+  public Object resolveArgument(MethodParameter methodParameter,
+                                ModelAndViewContainer modelAndViewContainer,
+                                NativeWebRequest nativeWebRequest,
+                                WebDataBinderFactory webDataBinderFactory) throws Exception {
+
+    val filters = new ArrayList<SearchFilter>();
+    nativeWebRequest.getParameterNames().forEachRemaining(p -> {
+      val matchingField = fieldValues.stream().filter(f -> f.equalsIgnoreCase(p)).findFirst();
+      if(matchingField.isPresent()){
+        filters.add(new SearchFilter(matchingField.get(),nativeWebRequest.getParameter(p)));
```
andricDu: Not a fan of a side effect two lambas deep but I'll admit converting an Iterator to a stream adds a lot of code clutter.

PR #55, opened 2017-11-28
add docker image

Jephuff:


cheapsteak:
Let's also add a .env.schema

DockerFile
```diff
@@ -0,0 +1,17 @@
...
+ADD pom.xml .
+RUN mvn verify clean --fail-never
+
+ADD . .
+
+EXPOSE 8081
+
+RUN mvn install -DskipTests
+
+CMD java -jar target/ego-0.0.1-SNAPSHOT-exec.jar --spring.datasource.url="$DB_URL" --spring.datasource.username="$DB_USER" --spring.datasource.password="$DB_PASSWORD"
```
joneubank: Part of the filename `0.0.1-SNAPSHOT` includes a version number that will change as releases are made. This command should work like this: `java -jar target/ego-*-exec.jar`

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: "3.2"
+services:
+  api:
+    build: .
+    ports:
+      - "$API_HOST_PORT:8081"
+    environment:
```
joneubank: Are we switching to a .env file instead?
If so, we should update that and include an example file.
Additional environment variables are needed than these three, take a look at the startup script used by CircleCI [start-server.sh](https://github.com/overture-stack/ego/blob/develop/src/main/resources/scripts/start-server.sh)

Jephuff: updated with the other variables and matched the names in the start script.
Also added a .env.schema file.

andricDu:
Thanks!

PR #58, opened 2017-11-29
Added JWT Signing with a key store and exposed the publick key endpoint

ra-ver:
JWT are now generated using key stored in a key store.
Exposed a public key end point. EGO applications can verify JWT using public key


joneubank:
For tracking purposes: https://github.com/overture-stack/ego/issues/28

Issue #68, opened 2018-01-03
Adding a user to a group from group page crashes page


rosibaj:
- Go to Group page
- click a group
- click edit
- click to add a user
- click on a user and the page will crash and just be blank instead of adding the user

PR #77, opened 2018-02-02
User.groups and User.applicaitons are now Sets

joneubank:
Fix situation where duplicate group/application additions to a user could cause issues. This is already protected against in the UI, but makes EGO enforce no duplicates as well.


joneubank:
https://github.com/overture-stack/ego/issues/68

Issue #74, opened 2018-02-02
Adding Application or Group removes the opposite values


rosibaj:
- got to a user that is assigned to a group
- click edit, add an application, and click save
- the application is saved, but all groups are removed
- the same happens in reverse when a user has applications, you add a group, applications will be removes

rosibaj:
@joneubank When i try and add a user to a group now, the page blanks out.  Is this related?

PR #79, opened 2018-02-05
UserService.update merges existing user, does not erase groups and apps

joneubank:
https://github.com/overture-stack/ego/issues/74

### DIFF
--- a/src/main/java/org/overture/ego/model/entity/User.java
+++ b/src/main/java/org/overture/ego/model/entity/User.java
@@ -137,7 +137,7 @@ public class User {
     }
   }
 
-  public void merge(User other) {
+  public void update(User other) {
     this.name = other.name;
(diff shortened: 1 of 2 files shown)


src/main/java/org/overture/ego/model/entity/User.java
```diff
@@ -137,4 +137,27 @@ protected void initGroups(){
...
 
+  public void update(User other) {
+    this.name = other.name;
+    this.firstName = other.firstName;
+    this.lastName = other.lastName;
+    this.role = other.role;
+    this.status = other.status;
+    this.preferredLanguage = other.preferredLanguage;
+
+    // Don't merge the ID, CreatedAt, or LastLogin date - those are procedural.
```
andricDu: 👍

PR #80, opened 2018-02-06
Apply Update pattern to Group and ApplicationApply update

joneubank:


joneubank:
Previous PR included the changes to User and UserService. For some reason those commits appear here again.

This PR applies the same changes to Group and Application.

PR #87, opened 2018-03-12
Fix User role update

ra-ver:
Implemented check for user role case sensitivity


ra-ver:
Not sure why this PR is showing JWT related changes too. That code is not touched in this PR and is already live in demo ego server.

andricDu:
Will merge pending checks.

PR #111, opened 2018-05-21
📌 Update maven version

dankolbman:
There must be a better way to do this?


ra-ver:
Thanks for updating this. We ran into it yesterday as well.

PR #139, opened 2018-08-21
Enhancement: We now always throw a DontSetIdInCreateRequestException …

KevinFHartmann:
…when an id that gets auto-generated is mistakenly provided in a POST request.

### DIFF
--- a/src/main/java/org/overture/ego/model/exceptions/DontSetIdInCreateRequestException.java
+++ b/src/main/java/org/overture/ego/model/exceptions/DontSetIdInCreateRequestException.java
@@ -1,0 +1,9 @@
+package org.overture.ego.model.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/overture/ego/model/exceptions/DontSetIdInCreateRequestException.java
```diff
@@ -0,0 +1,9 @@
+package org.overture.ego.model.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value= HttpStatus.BAD_REQUEST, reason=DontSetIdInCreateRequestException.reason)
+public class DontSetIdInCreateRequestException extends RuntimeException {
```
andricDu: can we rename this to `PostWithIdentifier`

KevinFHartmann: Sure! :-)

PR #149, opened 2018-08-28
Change: Removed ACL Controller, put ACL controller changes into Polic…

KevinFHartmann:
…yController

Change: We had two different things named "Permissions". Named the Permissions entity "Scope", and the specific permission setting for a Scope as a "Permission".
Bugfix: We no longer use lazy initialization for our table joins, so JSON can serialize it properly.


src/main/java/org/overture/ego/utils/AclPermissionUtils.java
```diff
@@ -6,8 +6,8 @@
 import java.util.stream.Collectors;
 
 public class AclPermissionUtils {
-  public static String extractPermissionString(Permission permission) {
-    return String.format("%s.%s", permission.getEntity().getName(), permission.getMask().toString());
+  public static String extractPermissionString(Permission scope) {
```
andricDu: why is it typed `permission` but named `scope`?

src/main/java/org/overture/ego/controller/PolicyController.java
```diff
@@ -28,7 +34,9 @@
 public class PolicyController {
 
   @Autowired
-  private AclEntityService aclEntityService;
+  private PolicyService policyService;
+  private GroupService groupService;
+  private UserService userService;
```
andricDu: You are missing autowired on the other services.

Lets use an autowired constructor instead of field injection. This will also allow us to declare our dependencies as final. 

```java
  private final PolicyService policyService;
  private final GroupService groupService;
  private final UserService userService;

  @Autowired
  public PolicyController(PolicyService policyService, GroupService groupService, UserService userService) {
    this.policyService = policyService;
    this.groupService = groupService;
    this.userService = userService;
  }
```

src/main/java/org/overture/ego/controller/PolicyController.java
```diff
@@ -91,29 +99,66 @@ Policy create(
...
-      @RequestHeader(value = HttpHeaders.AUTHORIZATION, required = true) final String accessToken,
-      @PathVariable(value = "id", required = true) String id) {
-    aclEntityService.delete(id);
+    @RequestHeader(value = HttpHeaders.AUTHORIZATION, required = true) final String accessToken,
+    @PathVariable(value = "id", required = true) String id) {
+    policyService.delete(id);
   }
+
+  public @ResponseBody
+  String createGroupPermission(
```
andricDu: what path is using?

src/main/java/org/overture/ego/model/entity/UserPermission.java
```diff
@@ -40,12 +40,12 @@
   UUID id;
 
   @NonNull
-  @ManyToOne(fetch = FetchType.LAZY)
+  @ManyToOne(fetch = FetchType.EAGER)
```
andricDu: why?

KevinFHartmann: This change fixes a bug with JSON serialization.

Issue #154, opened 2018-10-02
Deleting group deletes all users in group


andricDu:
`CascadeType.ALL` is being used where it shouldn't be.

Also applies to deleting applications where it will delete all associated users, groups, and users of those groups.

andricDu:
PR: #155

PR #155, opened 2018-10-03
Feature/cascading hibernate

andricDu:
Fixes the cascade type and move the cascade definition to use the hibernate annotations rather than the JPA ones. 

Added tests on group service to verify applications and groups will not get wiped out through a rogue cascade on delete by deleting a group. 

Added an ignored test for token services allowing quick stubbing and debugging of token generation.  

Related to this issue: #154

### DIFF
--- a/src/test/java/org/overture/ego/token/TokenServiceTest.java
+++ b/src/test/java/org/overture/ego/token/TokenServiceTest.java
@@ -63,7 +63,6 @@ public class TokenServiceTest {
     val group2 = groupService.getByName("testGroup");
     group2.addUser(user);
     groupService.update(group2);
-    val groupShit = groupService.getByName("testGroup");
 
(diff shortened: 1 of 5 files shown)


src/test/java/org/overture/ego/token/TokenServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    val app2 = applicationService.getByClientId("foo");
+    app2.setWholeUsers(Sets.newHashSet(user));
+    applicationService.update(app2);
+
+
+    val token = tokenService.generateUserToken(userService.get(user.getId().toString()));
+  }
+
+}
```
lepsalex: **cough **cough

https://robots.thoughtbot.com/no-newline-at-end-of-file

PR #185, opened 2018-12-03
Bug fix for updating fields on a group.

andricDu:
Makes the update method return a new group rather than being a void method.


src/main/java/bio/overture/ego/model/entity/Group.java
```diff
@@ -62,12 +64,12 @@
     strategy = "org.hibernate.id.UUIDGenerator")
   @GeneratedValue(generator = "group_uuid")
   UUID id;
-  @Column(nullable = false, name = Fields.NAME, updatable = false)
+  @Column(nullable = false, name = Fields.NAME, updatable = true)
```
joneubank: If updatable=true is default, just remove from annotation.

PR #219, opened 2018-12-12
Catch jwt parse error

d8660091:


src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -227,8 +227,12 @@ public String generateAppToken(Application application) {
   }
 
   public boolean validateToken(String token) {
```
andricDu: Can we rename this to `isValidToken`

PR #240, opened 2019-01-30
Implemented admin role for applications

KevinFHartmann:
Enhancement: Ego applications now require a field called role, with v…alues 'USER' or 'ADMIN'.

Enhancement: Ego now grants the ADMIN role when the Authorization field
contains either a user authentication token for a user who has role ADMIN
(ie.'Bearer'<space><EgoUserJWT> (obtained from the checkToken endpoint)),
OR an application authentication token for an appliacation with role ADMIN
(ie.'Basic'<space><base64encoded <clientid>':'<clientsecret>>)'.


src/main/java/bio/overture/ego/model/enums/JavaFields.java
```diff
@@ -26,7 +26,7 @@
   public static final String ID = "id";
   public static final String NAME = "name";
   public static final String EMAIL = "email";
-  public static final String ROLE = "role";
+  public static final String ROLE = "type";
```
andricDu: why is this constant still named `ROLE`?

andricDu:
@KevinFHartmann looks like there is a merge conflict.

PR #245, opened 2019-02-04
Create an endpoint to list tokens for given user.

Buwujiu:


src/main/java/bio/overture/ego/controller/TokenController.java
```diff
@@ -98,9 +98,12 @@ public TokenController(@NonNull TokenService tokenService) {
     return format("Token '%s' is successfully revoked!", token);
   }
 
-  @ResponseBody
-  List<TokenResponse> listTokens(@RequestHeader(value = "Authorization") String authorization) {
-    return null;
+  @RequestMapping(method = RequestMethod.GET, value = "/list_token")
```
andricDu: lets rename this to just token.

This way we get proper REST usaged with GET, POST, and DELTE for /token.

I realize now the check_token endpoint must be confusing you. The reason that exists is for backwards compatibility with ICGC.

andricDu: I think you are missing the annotation `@AdminScoped` to make sure only an admin user or application can list this.

Buwujiu:
ticket #245

andricDu:
Thanks!

PR #284, opened 2019-03-01
revokeToken endpoint changes.

Buwujiu:


src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -345,23 +351,39 @@ public TokenScopeResponse checkToken(String authToken, String token) {
...
+  }
 
-    log.info(format("validating if user '%s' has permission to revoke token.", str(userId)));
-    if (userService.isAdmin(user) && userService.isActiveUser(user)) {
-      revokeToken(tokenName);
+  private void revokeTokenAsUser(String tokenName) {
+    val token =
+        findByTokenString(tokenName)
+            .orElseThrow(() -> new InvalidTokenException("Token not found! "));
+    if (userService.isAdmin(token.getOwner()) && userService.isActiveUser(token.getOwner())) {
```
andricDu: We should be checking if the current principal (requesting user) is an admin, not the token owner.

andricDu:
Thanks!

PR #309, opened 2019-03-18
301 revoke tokens on permission

andricDu:
Tokens revoked correctly when resolved scopes for a user change and make existing tokens invalid.


andricDu:
Two changes:
1. On group deletes, verify tokens of users who were part of group
2. On user delete, revoke tokens

andricDu:
@rtisma does it make sense to cascade deletes of users to their tokens?

PR #313, opened 2019-03-19
Removes association between apps and api tokens

andricDu:


src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -333,29 +323,20 @@ public TokenScopeResponse checkToken(String authToken, String token) {
       throw new InvalidTokenException("No token field found in POST request");
     }
 
-    log.info(format("token ='%s'", token));
+    log.debug(format("token ='%s'", token));
```
rtisma: we should Md5 this. tokens in a log file arent the best idea

PR #315, opened 2019-03-20
209 refactor last login to remove reactor

KevinFHartmann:


src/main/java/bio/overture/ego/service/AbstractPermissionService.java
```diff
@@ -13,14 +13,14 @@
...
 import static com.google.common.collect.Maps.uniqueIndex;
-import static com.gs.collections.impl.factory.Sets.intersect;
 import static java.util.Arrays.stream;
 import static java.util.Collections.reverse;
 import static java.util.Comparator.comparing;
 import static java.util.Objects.isNull;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.toMap;
+import static org.eclipse.collections.impl.factory.Sets.intersect;
```
andricDu: We don't need eclipse libraries to figure out intersections. We can use google guava which we already have on the classpath. 

```java
import com.google.common.collect.Sets;
```
And then later:

```java
Sets.intersection(...)
```

pom.xml
```diff
@@ -235,16 +235,17 @@
...
         </dependency>
 
-        <!-- Reactor -->
+        <!-- Eclipse collections --> 
         <dependency>
-            <groupId>io.projectreactor</groupId>
-            <artifactId>reactor-bus</artifactId>
-            <version>2.0.8.RELEASE</version>
+            <groupId>org.eclipse.collections</groupId>
+            <artifactId>eclipse-collections-api</artifactId>
```
andricDu: See my other comment about set intersection.

PR #320, opened 2019-03-22
Use one single save instead of multiple in entity generator

d8660091:


andricDu:
@d8660091 merge conflicts

andricDu:
@d8660091 merge conflicts!

Issue #331, opened 2019-04-04
Token Unit Tests - Data Integrity Broken


andricDu:
Just don't reuse names for tokens.

andricDu:
https://github.com/overture-stack/ego/pull/332

andricDu:
Should have been closed a long time ago.

PR #332, opened 2019-04-04
fixes unit tests by not re-using same token identifiers

andricDu:


andricDu:
https://github.com/overture-stack/ego/issues/331

PR #355, opened 2019-05-27
Change user proto to return list of strings

joneubank:
Changing lists of Boxed variables to list of primitives.

### DIFF
--- a/src/main/proto/Ego.proto
+++ b/src/main/proto/Ego.proto
@@ -53,7 +53,7 @@ message User {
     google.protobuf.StringValue status = 9;
     google.protobuf.StringValue type = 10;
 
-    repeated google.protobuf.StringValue applications = 11;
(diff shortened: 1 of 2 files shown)


src/main/proto/Ego.proto
```diff
@@ -53,7 +53,7 @@ message User {
     google.protobuf.StringValue status = 9;
     google.protobuf.StringValue type = 10;
 
-    repeated google.protobuf.StringValue applications = 11;
-    repeated google.protobuf.StringValue groups = 12;
-    repeated google.protobuf.StringValue scopes = 13;
+    repeated string applications = 11;
+    repeated string groups = 12;
+    repeated string scopes = 13;
```
rtisma: might be more descriptive to put application_ids, group_ids, scopes.

PR #370, opened 2019-08-01
Fix/test compile time

andricDu:
### Reason
Post upgrade to JDK11, compile time for tests was taking far too long to be reasonable. I suspected AssertJ was to blame due to some sort of high complexity type inference analysis cause by the "fluent" API crashing against generics. 

### Changes
- Removed explicit AssertJ dependency as it's already provided by Spring Boot Test Start
- Updated Spring Boot version
- Removed AssertJ usage from several files to improve test compile time.

### Results
```bash
mvn clean test-compile
```

With AssertJ `assertThat`
```
[INFO] Total time:  04:11 min
```

With JUnit `Assert`
```
[INFO] Total time:  39.023 s
```

### Extra Notes
AssertJ has not been completely removed but it most probably should. In the future we should see if we can use JUnit5 which has better support for testing exceptions which is the real **functional** benefit AssertJ provided.


src/test/java/bio/overture/ego/controller/ApplicationControllerTest.java
```diff
@@ -117,9 +115,9 @@ public void addApplication_Success() {
     val response = initStringRequest().endpoint("/applications").body(app).post();
 
     val responseStatus = response.getStatusCode();
-    assertThat(responseStatus).isEqualTo(HttpStatus.OK);
+    Assert.assertEquals(responseStatus, HttpStatus.OK);
```
KevinFHartmann: Make these static imports.

src/test/java/bio/overture/ego/controller/UserControllerTest.java
```diff
@@ -207,32 +204,28 @@ public void createUser_NonExisting_Success() {
...
+    Assert.assertEquals(r1.getPreferredLanguage(), r.getPreferredLanguage());
+    Assert.assertEquals(r1.getType(), r.getType());
+    Assert.assertEquals(r1.getStatus(), r.getStatus());
+
+    Assert.assertEquals(r1.getEmail(), user.getEmail());
+    Assert.assertEquals(r1.getFirstName(), user.getFirstName());
+    Assert.assertEquals(r1.getLastName(), user.getLastName());
+    Assert.assertEquals(r1.getPreferredLanguage(), user.getPreferredLanguage());
+    Assert.assertEquals(r1.getType(), user.getType());
+    Assert.assertEquals(r1.getStatus(), user.getStatus());
```
KevinFHartmann: The problem with comparing specific fields is that if one is added, you have to remember to update the test.

If that's really an issue, though, we can do something with reflection (if we have to...)

andricDu: True, but right now I believe it to be one of the biggest hits on compile time. What might be good is if we just implement it ourselves at some point on top of JUnit

PR #380, opened 2019-08-20
Token Creation self service

andricDu:
## Feature Changes to Access Token Creation
- Admin Users **CAN** create Access Tokens on behalf of other users (JWT Bearer)
- Admin Application **CAN** create Access Tokens on behalf of users (JWT Bearer ***OR*** Basic Auth)
- Non-Admin Users **CAN** create Access Tokens for themselves (JWT Bearer)
- Non-Admin Users **CANNOT** create Access Tokens on behalf of others
- Non-Admin Applications **CANNOT** create Access Tokens on behalf of others

## Code Changes
- Auth check added to `POST /o/token` controller as described above.
- Unit Tests to validate behavior as described above
- Updated EntityGenerator to be able to create regular users in addition to admin users
- Update broken test that was previously returning 500 but with changes now returns 403

### DIFF
--- a/src/main/java/bio/overture/ego/controller/TokenController.java
+++ b/src/main/java/bio/overture/ego/controller/TokenController.java
@@ -32,9 +32,12 @@ import bio.overture.ego.model.dto.Scope;
 import bio.overture.ego.model.dto.TokenResponse;
 import bio.overture.ego.model.dto.TokenScopeResponse;
 import bio.overture.ego.model.dto.UserScopesResponse;
+import bio.overture.ego.model.entity.User;
+import bio.overture.ego.model.exceptions.ForbiddenException;
 import bio.overture.ego.model.params.ScopeName;
 import bio.overture.ego.security.AdminScoped;
 import bio.overture.ego.security.ApplicationScoped;
+import bio.overture.ego.security.AuthorizationManager;
 import bio.overture.ego.service.TokenService;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/overture/ego/controller/TokenController.java
```diff
@@ -101,6 +110,21 @@ public TokenController(@NonNull TokenService tokenService) {
...
       @RequestParam(value = "description", required = false) String description) {
+
+    // Check if admin, if not, then check if owner
+    val authentication = SecurityContextHolder.getContext().getAuthentication();
+    if (!authorizationManager.authorizeWithAdminRole(authentication)) {
+      val principal = authentication.getPrincipal();
+      if (principal instanceof User) {
+        User user = (User) principal;
+        if (!user.getId().equals(user_id)) {
+          throw new ForbiddenException("Forbidden");
```
joneubank: Does this message get used? Could be more helpful ("Users cannot create tokens for other users")

src/main/java/bio/overture/ego/controller/TokenController.java
```diff
@@ -101,6 +110,21 @@ public TokenController(@NonNull TokenService tokenService) {
...
+        }
+      } else {
+        throw new ForbiddenException("Forbidden");
```
joneubank: Same thought about exception messaging.

src/main/java/bio/overture/ego/controller/TokenController.java
andricDu: I was under the impression that in the case a user is performing an action that is `Forbidden` you want to avoid giving them an indication as to the details of why. Could I satisfy you with a log?

PR #403, opened 2019-10-15
Ego authentication doesnt verify the client secret for applications#402

KevinFHartmann:


src/main/java/bio/overture/ego/security/BasicAuthToken.java
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      contents = new String(Base64.getDecoder().decode(base64encoding));
+    } catch (Exception exception) {
+      log.error("Couldn't decode basic auth token: '" + token + "', " + exception.getMessage());
+      return Optional.empty();
+    }
+    val parts = COLON_SPLITTER.splitToList(contents);
+    if (parts.size() != 2) {
+      log.error("Basic auth token '" + token + "' should have 2 parts, not " + parts.size());
+      return Optional.empty();
```
andricDu: Optionals!!! :+1:

src/main/java/bio/overture/ego/security/JWTAuthorizationFilter.java
```diff
@@ -127,13 +127,45 @@ private void authenticateUserOrApplication(String tokenPayload) {
...
+    }
+
+    // Deny access if the clientSecret in the token is wrong
+    if (!application.get().getClientSecret().equals(clientSecret)) {
+      SecurityContextHolder.clearContext();
+      log.warn(
+          "AuthenticateApplication: Wrong client secret for clientId '"
+              + clientId
+              + "' from token '"
+              + tokenString
```
andricDu: lets avoid putting the whole token string into logs as it contains the secret.

andricDu: If you feel strongly we should be logging the token for debugging purposes then maybe add a separate log at DEBUG level.

src/main/java/bio/overture/ego/security/JWTAuthorizationFilter.java
```diff
@@ -69,7 +69,7 @@ public void doFilterInternal(
     }
     val tokenPayload = request.getHeader(HttpHeaders.AUTHORIZATION);
 
-    if (tokenPayload != null && tokenPayload.startsWith(applicationService.APP_TOKEN_PREFIX)) {
+    if (tokenPayload != null && tokenPayload.startsWith(BasicAuthToken.TOKEN_PREFIX)) {
```
andricDu: This is incorrect as this turns Basic tokens in Bearer tokens. APP_TOKEN_PREFIX is declared as the following in the ApplicationService class:
```java
public static final String APP_TOKEN_PREFIX = "Basic ";
```

This would break the client credentials OAuth2 flow as Basic auth would be rejected.

I just verified and the tests pass which is scary given this particular line, so lets fix this and add a test for that. 
The test should do the following:
 1. Have an app registered to ego
 2. Use client credentials with basic token to get application JWT.
 3. Verify that is was successful

andricDu: ignore me

PR #404, opened 2019-10-15
Add front_state parameter

d8660091:
This commit allows frontend to pass state to ego. It can be used to handle join program state in the url parameter `https://ego_api/oauth/login/google?client_id=platform-ui-qa&front_state=join_program`, and the frontend can get the state back to redirect the user to the correct page

### DIFF
--- a/src/main/java/bio/overture/ego/security/OAuth2ClientResources.java
+++ b/src/main/java/bio/overture/ego/security/OAuth2ClientResources.java
@@ -24,10 +24,17 @@ public class OAuth2ClientResources {
             val uri = new URI(request.getCurrentUri());
             val attr = (ServletRequestAttributes) RequestContextHolder.currentRequestAttributes();
             val session = attr.getRequest().getSession(true);
-            val pattern = Pattern.compile("client_id=([^&]+)?");
-            val matcher = pattern.matcher(uri.getQuery());
-            if (matcher.find()) {
-              session.setAttribute("ego_client_id", matcher.group(1));
+
+            val clientIdPattern = Pattern.compile("client_id=([^&]+)?");
+            val clientIdMatcher = clientIdPattern.matcher(uri.getQuery());
+            if (clientIdMatcher.find()) {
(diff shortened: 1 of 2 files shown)


andricDu:
### Reasons why this PR is being rejected in order of importance
1. This PR breaks with what is described in the ticket **without a discussion with the team** to see if this change is an appropriate technical solution to deliver the feature. 
   - Described here: https://github.com/icgc-argo/platform-ui/issues/594 `Register a 2nd ego application for the login page`
   - Good reasons were given by the team for the agreed implementation. Business logic out of UI code, avoid global state bugs in UI, handling of redirection in backend as good practice. 
2. This implementation touches ego when the original discussion did not call for it **nor is there an issue to track** a change made to ego. The **owners of the ego code base were not consulted** if this change should be made. 
3. This implementation implies certain changes will need to be made to the platform ui **without consulting the owners of the platform ui.** 
4. The query param `front_state` is poorly named which seems to indicate that **this change was made as a quick fix**, which the lack of tests for such an important feature seem to confirm. Critical changes around security should not be made hastily and there should be tests to verify functionality. 
5. **Your implementation is incorrect** as its not part of the OAuth2.0 spec. The correct way to implement something to the effect of what you were aiming for would be to add the OAuth2.0 Implicit flow which has an aptly named `redirect_uri` query param already part of the spec. 
    - https://tools.ietf.org/html/rfc6749#section-1.3.2

**Take note, the most problematic issue is the lack of communication and collaboration with the rest of the team.  The fact about your implementation being wrong is the smallest problem.**

d8660091:
@andricDu 
Although it's closed, but I felt it's worth to address your comment:
1. The content of the pull request has not been discussed in the team. I felt itself makes a good place for discussion. We can reopen the issue and have some discussion on it directly.

2,3 Pull request is a discussion place for whoever including the ego code contributors.

5. Your argument is wrong. The communication between ego and google is oauth2, but the same doesn't apply to the communication between ego and ego client.

PR #423, opened 2019-11-25
Allows clients to specify redirect_uri query param.

andricDu:
## Summary
Provides support for clients to provide a redirect URI as a query param `redirect_uri`.  Applications should register a comma-delimited list of redirect URIs in the redirect uri fields. Redirect URIs just have to match the same authority (domain and port) and can be of dynamic path. If no redirect URI is present, the first valid URI from the list is used verbatim. 

## Changes
- new query param `redirect_uri` added to login flow
- persistence of redirection to session nonce
- CORS validation with redirect uri 
- Utility class `Redirects` for correctly matching a redirect uri and verifying validity
- Unit tests
- Removed unused badges from README

### DIFF
--- a/src/main/java/bio/overture/ego/security/OAuth2ClientResources.java
+++ b/src/main/java/bio/overture/ego/security/OAuth2ClientResources.java
@@ -2,7 +2,7 @@ package bio.overture.ego.security;
 
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.regex.Pattern;
+import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.springframework.boot.autoconfigure.security.oauth2.resource.ResourceServerProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
@@ -10,7 +10,9 @@ import org.springframework.security.oauth2.client.token.AccessTokenRequest;
 import org.springframework.security.oauth2.client.token.grant.code.AuthorizationCodeResourceDetails;
 import org.springframework.web.context.request.RequestContextHolder;
 import org.springframework.web.context.request.ServletRequestAttributes;
+import org.springframework.web.util.UriComponentsBuilder;
 
+@Slf4j
 public class OAuth2ClientResources {
 
   @NestedConfigurationProperty
@@ -24,10 +26,20 @@ public class OAuth2ClientResources {
             val uri = new URI(request.getCurrentUri());
             val attr = (ServletRequestAttributes) RequestContextHolder.currentRequestAttributes();
             val session = attr.getRequest().getSession(true);
-            val pattern = Pattern.compile("client_id=([^&]+)?");
-            val matcher = pattern.matcher(uri.getQuery());
-            if (matcher.find()) {
-              session.setAttribute("ego_client_id", matcher.group(1));
+
+            val uriComponents = UriComponentsBuilder.fromUri(uri).build();
+            val queryParams = uriComponents.getQueryParams();
+
+            val clientId = queryParams.getFirst("client_id");
+            if (clientId != null && !clientId.isEmpty()) {
+              session.setAttribute("ego_client_id", clientId);
+            }
+
+            val redirectUri = queryParams.getFirst("redirect_uri");
+            if (redirectUri != null && !redirectUri.isEmpty()) {
+              session.setAttribute("ego_redirect_uri", redirectUri);
+            } else {
+              session.setAttribute("ego_redirect_uri", "");
             }
 
             if (getPreEstablishedRedirectUri() != null) {
--- a/src/main/java/bio/overture/ego/utils/Redirects.java
+++ b/src/main/java/bio/overture/ego/utils/Redirects.java
@@ -43,7 +43,7 @@ public class Redirects {
     val redirects = Arrays.stream(app.getRedirectUri().split(","));
 
     // Short return if no redirect URI is provided
-    if (redirect == null || redirect.isEmpty()) {
+    if (redirect == null || redirect.isBlank()) {
       return redirects
           .findFirst()
           .orElseThrow(() -> new UnauthorizedClientException("Cannot find valid redirect URI"));
--- a/src/main/java/bio/overture/ego/utils/Redirects.java
+++ b/src/main/java/bio/overture/ego/utils/Redirects.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2019. The Ontario Institute for Cancer Research. All rights reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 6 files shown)


src/main/java/bio/overture/ego/utils/Redirects.java
```diff
@@ -0,0 +1,89 @@
...
+          .orElseThrow(() -> new UnauthorizedClientException("Cannot find valid redirect URI"));
+    }
+
+    val msg = format("Unauthorized redirect URI: %s", redirect);
+    URI redirectUri;
+    try {
+      redirectUri = new URI(redirect);
+    } catch (URISyntaxException e) {
+      log.error(msg);
+      throw new UnauthorizedClientException(msg);
```
blabadi: are those unauthorized or bad request ?

src/main/java/bio/overture/ego/utils/Redirects.java
```diff
@@ -0,0 +1,89 @@
...
+                        e);
+                    return null;
+                  }
+                })
+            .filter(Objects::nonNull)
+            .map(
+                u ->
+                    u.getHost().equals(redirectUri.getHost())
+                        && u.getPort() == redirectUri.getPort()) // Map to valid/invalid
+            .reduce(Boolean::logicalOr) // Needs at least one valid
```
blabadi: you could use `anyMatch` if that makes it cleaner

src/main/java/bio/overture/ego/utils/Redirects.java
```diff
@@ -0,0 +1,89 @@
...
+      redirectUri = new URI(redirect);
+    } catch (URISyntaxException e) {
+      log.error(msg);
+      throw new UnauthorizedClientException(msg);
+    }
+
+    val isValid =
+        redirects
+            .map(
+                r -> {
```
blabadi: use a private function if possible

src/main/java/bio/overture/ego/utils/Redirects.java
andricDu: I'd argue unauthorized as by the time it gets here, I've understood the request.

src/main/java/bio/overture/ego/security/OAuth2ClientResources.java
```diff
@@ -24,10 +26,20 @@ public String getRedirectUri(AccessTokenRequest request) {
...
-            val matcher = pattern.matcher(uri.getQuery());
-            if (matcher.find()) {
-              session.setAttribute("ego_client_id", matcher.group(1));
+
+            val uriComponents = UriComponentsBuilder.fromUri(uri).build();
+            val queryParams = uriComponents.getQueryParams();
+
+            val clientId = queryParams.getFirst("client_id");
+            if (clientId != null && !clientId.isEmpty()) {
+              session.setAttribute("ego_client_id", clientId);
```
blabadi: since this and the redirect uri are extracted from the url, do we really need to store them in a session instead of extracting them again before redirect assuming we have access to the original request object ? im a bit ignorant on the redirection dance, so asking more for my information

andricDu: I believe its being used this way because the implementation isn't using the state parameter from the oauth2 spec correctly.

src/main/java/bio/overture/ego/utils/Redirects.java
andricDu: wow that makes it much cleaner!

src/main/java/bio/overture/ego/utils/Redirects.java
andricDu: :+1:

src/main/java/bio/overture/ego/security/OAuth2ClientResources.java
andricDu: To be explicit, instead of storing this in the session, we would put the clientid and redirect into a JSON object that we also make as a nonce and we base64 encode and put as the state.

andricDu: Basically have something like this:
```java
restTemplate.getOAuth2ClientContext().setPreservedState(ourStateInsteadOfSession);
```
in this class: https://github.com/overture-stack/ego/blob/develop/src/main/java/bio/overture/ego/security/OAuth2ClientResources.java

src/main/java/bio/overture/ego/utils/Redirects.java
```diff
@@ -0,0 +1,88 @@
...
+   * Returns redirect uri based on combination between provided URI and those registed in the passed
+   * application. Checks validity and throws 403 if unauthorized URI is provided.
+   *
+   * @return Returns URI as a String that Ego will redirect to
+   */
+  public static String getRedirectUri(@NonNull Application app, String redirect) {
+    val redirects = Arrays.stream(app.getRedirectUri().split(","));
+
+    // Short return if no redirect URI is provided
+    if (redirect == null || redirect.isEmpty()) {
```
blabadi: edge case : non empty, whitespace string

andricDu: Yea, I thought about this, whether or not I should do stripping after the splitting. 

My logic is if we are already in the land of comma delimited fields, it's a game of whack-a-mole with the classes of bugs that brings so I decided to not hold the hand of the user/client that sets up an application.

I can revisit this if you feel strongly otherwise.

andricDu: ignore what I just said, that is legit.

andricDu: switch to `isBlank`

src/test/java/bio/overture/ego/utils/RedirectsTest.java
```diff
@@ -0,0 +1,109 @@
...
+import static org.junit.Assert.*;
+
+import bio.overture.ego.model.entity.Application;
+import lombok.val;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.springframework.security.oauth2.common.exceptions.UnauthorizedClientException;
+
+public class RedirectsTest {
```
hlminh2000: should we have a test case with url-encoded query params? something like `http://example.com/path?query=123`

andricDu: yes, I'll add

blabadi: but with the current matching strategy (host+port) this won't guarantee you will get the correct path.. if you have: `http://example.com/path?query=123` & `http://example.com/path2`

andricDu: What do you mean? If you provide a redirect uri from the client, any registered uri just needs to match the authority(domain, port) for the provided uri to be accepted

blabadi: yeah my bad, I just confused myself

andricDu: :spaghetti: no worries, it is confusing as its supporting two modes of operation at the same time :spaghetti:

PR #424, opened 2019-11-25
Rtisma test refresh token setup

rtisma:
2 commits. ignore the "formatted" one. maven automattically formats the code. The commit with "fix" represents the changes to make it all work


andricDu:
@rtisma can this be closed?

PR #427, opened 2019-11-28
updated

rtisma:


src/main/java/bio/overture/ego/service/RefreshTokenService.java
```diff
@@ -27,18 +31,37 @@
...
     this.refreshTokenRepository = refreshTokenRepository;
   }
 
+  //TODO: Ann to implement this
+  private RefreshToken  createTransient(User user){
+    val nowInstant = Instant.now();
+   val expiryInstant = nowInstant.plus(durationSeconds, SECONDS);
+    return RefreshToken.builder()
+        .jti(UUID.randomUUID()) //TODO [ANN] to implement propertly with jwt
+        .issueDate(Date.from(nowInstant))
```
anncatton: are `issueDate` and `expiryDate` moved here instead of the db to make them easier to manage?

src/test/java/bio/overture/ego/utils/EntityGenerator.java
```diff
@@ -493,11 +519,9 @@ private static String generateRandomUserName(Random r, int length) {
     return fn + " " + ln;
   }
 
-  public RefreshToken setupRefreshToken(User user) {
+  public RefreshToken generateRefreshToken(@NonNull User user) {
```
anncatton: i guess this can get removed now the service method is there :)

src/main/java/bio/overture/ego/model/entity/RefreshToken.java
```diff
@@ -54,11 +64,13 @@
 //  private User user;
 
   @NotNull
-  @Column(name = SqlFields.ISSUEDATE, updatable = false, nullable = false)
+  @Column(name = SqlFields.ISSUEDATE)
+  @Temporal(value = TemporalType.TIMESTAMP)
```
anncatton: is this annotation just a simple way to match the field with the db type?

src/main/java/bio/overture/ego/service/RefreshTokenService.java
```diff
@@ -27,18 +31,37 @@
...
 
   @Autowired
   public RefreshTokenService(
-    @NonNull RefreshTokenRepository refreshTokenRepository,
-    @NonNull UserRepository userRepository,
-    @NonNull UserService userService) {
+      @NonNull RefreshTokenRepository refreshTokenRepository,
+      @NonNull UserRepository userRepository,
+      @NonNull UserService userService,
+      @Value("${refreshToken.durationSeconds:10800000}") int durationSeconds
```
anncatton: does this value refer to what's in the `application.yml`? if it is why does it need the seconds value added? i may be thinking of the yml file incorrectly...i'm looking at the `refreshToken.durationSeconds` as a constant...

src/main/java/bio/overture/ego/service/RefreshTokenService.java
```diff
@@ -27,18 +31,37 @@
...
+  ) {
     super(RefreshToken.class, refreshTokenRepository);
+    this.durationSeconds = durationSeconds;
     this.refreshTokenRepository = refreshTokenRepository;
   }
 
+  //TODO: Ann to implement this
+  private RefreshToken  createTransient(User user){
```
anncatton: does the method name here imply it should use the `transient` keyword too?

andricDu:
@rtisma @anncatton can this be closed?

anncatton:
oh yeah, forgot about this one - we can close it!

PR #431, opened 2019-12-10
394 implement refresh token

anncatton:
- cleans up logic for expired refresh token
- cleans up comments

### DIFF
--- a/src/test/java/bio/overture/ego/service/RefreshContextServiceTest.java
+++ b/src/test/java/bio/overture/ego/service/RefreshContextServiceTest.java
@@ -1,5 +1,7 @@
 package bio.overture.ego.service;
 
+import static bio.overture.ego.model.enums.StatusType.PENDING;
+
 import bio.overture.ego.model.domain.RefreshContext;
 import bio.overture.ego.model.exceptions.ForbiddenException;
 import bio.overture.ego.model.exceptions.NotFoundException;
@@ -19,7 +21,6 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.transaction.annotation.Transactional;
-import static bio.overture.ego.model.enums.StatusType.PENDING;
 
 @Slf4j
 @SpringBootTest
@@ -44,7 +45,9 @@ public class RefreshContextServiceTest {
     val user1Token = tokenService.generateUserToken(user1);
     val refreshToken1 = user1.getRefreshToken();
 
+    Assert.assertTrue(refreshContextService.findById(refreshToken1.getId()).isPresent());
     Assert.assertTrue(refreshContextService.getById(refreshToken1.getId()) != null);
+
     val incomingRefreshContext =
         refreshContextService.createRefreshContext(refreshToken1.getId().toString(), user1Token);
     refreshContextService.disassociateUserAndDelete(user1Token);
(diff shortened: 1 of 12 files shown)


src/test/java/bio/overture/ego/service/RefreshContextServiceTest.java
```diff
@@ -36,11 +37,34 @@
   @Rule public ExpectedException exceptionRule = ExpectedException.none();
 
   @Test
-  public void refresh_usedTokenIsDeleted() {
+  public void refresh_invalidContext_usedTokenIsDeleted() {
+    val user1 = entityGenerator.setupUserWithRefreshToken("User One");
+    // force invalid refreshContext with invalid user status
+    user1.setStatus(StatusType.PENDING);
```
rtisma: Our general guideline for static methods/fields is to statically import them. Intellij has an option that can do the static import for you

src/test/java/bio/overture/ego/service/RefreshContextServiceTest.java
```diff
@@ -36,11 +37,34 @@
...
+    val user1Token = tokenService.generateUserToken(user1);
+    val refreshToken1 = user1.getRefreshToken();
+
+    Assert.assertTrue(refreshContextService.getById(refreshToken1.getId()) != null);
```
rtisma: this assertion is not enough. A better one is:
```java
assertTrue(refreshContextService.getById(refreshToken1.getId) != null);
assertTrue(refreshContextService.getById(refreshToken1.getId()).isPresent());
```
Even just doing the isPresent check is enough for a test. In general, if an `Optional<>` object is returned then you can assume the developer intends it to always be not null. you can have empty optional objects, and non-empty optional objects. Returning null on top of that defeats the purpose of Optional :)

src/test/java/bio/overture/ego/service/RefreshContextServiceTest.java
```diff
@@ -36,11 +37,34 @@
...
+    val incomingRefreshContext =
+        refreshContextService.createRefreshContext(refreshToken1.getId().toString(), user1Token);
+    refreshContextService.disassociateUserAndDelete(user1Token);
+
+    exceptionRule.expect(NotFoundException.class);
+    exceptionRule.expectMessage(
+        String.format("RefreshToken '%s' does not exist", refreshToken1.getId()));
+    Assert.assertTrue(refreshContextService.get(refreshToken1.getId(), false) == refreshToken1);
```
rtisma: this is a java gotcha! Read below
https://stackoverflow.com/a/10806192/1952450

basically, you have to use `.equals()` for this type of comparison

In a nut shell, the following code runs without failure/errors:
```java
RefreshToken r1 = RefreshToken.builder().id("someId").build();
RefreshToken r2 = r1;
RefreshToken r3 = RefreshToken.builder().id("someId").build();
RefreshToken r4 = RefreshToken.builder().id("SOMETHING_ELSE").build();

assertTrue(r1 == r2);
assertTrue(r1 != r3);
assertTrue(r1 != r4);
assertTrue(r2 != r3);
assertTrue(r2 != r4); 
assertTrue(r3 != r4);

assertTrue(r1.equals(r2));
assertTrue(r1.equals(r3));
assertFalse(r1.equals(r4));
assertTrue(r2.equals(r3));
assertFalse(r2.equals(r4));
assertFalse(r3.equals(r4));
```

For the selected line of code, you can use:
```java
assertEquals(refreshContextService.get(refreshToken1.getId(), false), refreshToken1);
```
which calls the .equals method under the hood.

anncatton: Aha...i ran into this with a string comparison but didn't realize it here!

anncatton:
ohh yes forgot about the assert imports - will do that before merging!

PR #432, opened 2019-12-11
Refresh token feature

anncatton:
Feature branch for refresh token - wip!


src/main/java/bio/overture/ego/service/RefreshContextService.java
```diff
@@ -0,0 +1,169 @@
...
+    return refreshTokenRepository.save(refreshToken);
+  }
+
+  public RefreshContext createRefreshContext(String refreshTokenId, String bearerToken) {
+    val refreshTokenOpt = refreshTokenRepository.findById(UUID.fromString(refreshTokenId));
+    val tokenClaims = tokenService.getTokenClaims(bearerToken);
+    val user = tokenService.getTokenUserInfo(bearerToken);
+
+    checkNotFound(
+        !isNull(refreshTokenOpt) && refreshTokenOpt.isPresent(),
```
andricDu: You don't need `!isNull(refreshTokenOpt)` in this check, as refreshTokenOpt is an optional type, it will never be null.

src/main/java/bio/overture/ego/service/RefreshContextService.java
```diff
@@ -0,0 +1,169 @@
...
+    createRefreshToken(newUserToken);
+    return newUserToken;
+  }
+
+  private Cookie createCookie(String cookieName, String cookieValue, Integer maxAge) {
+    Cookie cookie = new Cookie(cookieName, cookieValue);
+    // where to access the accepted domain?
+    cookie.setDomain("localhost");
+    // disable setSecure while testing locally in browser, or will not show in cookies
+    //    cookie.setSecure(true);
```
andricDu: lets make this a configuration value in the application.yml

That way in the dev/test profile we can have this set to false.

src/main/java/bio/overture/ego/service/RefreshContextService.java
```diff
@@ -0,0 +1,169 @@
...
+        .build();
+  }
+
+  public Optional<RefreshToken> getRefreshTokenByUser(User user) {
+    return refreshTokenRepository.getByUser(user);
+  }
+
+  private void checkUniqueByUser(User user) {
+    checkUnique(
+        !getRefreshTokenByUser(user).isPresent(),
```
andricDu: avoid negating during conditionals if you have access to a Boolean method that satisfies the semantics and can use it without negating. 

```java
getRefreshTokenByUser(user).isEmpty(),
```

anncatton:
I'll add the changes for the expired jwt in a PR and attach https://github.com/overture-stack/ego/issues/434

andricDu:
LGTM!

PR #436, opened 2020-01-03
Add pagination, update apiKey GET response

anncatton:


src/main/java/bio/overture/ego/repository/queryspecification/TokenStoreSpecification.java
```diff
@@ -16,6 +16,34 @@
...
+import bio.overture.ego.utils.QueryUtils;
+import java.util.UUID;
+import javax.persistence.criteria.Join;
+import lombok.NonNull;
+import lombok.val;
+import org.springframework.data.jpa.domain.Specification;
+
+public class TokenStoreSpecification extends SpecificationBase<ApiKey> {
+
+  public static Specification<ApiKey> containsText(@NonNull String text) {
```
andricDu: Maybe I'm misunderstanding, but wouldn't this be clearer if this method was called `containsName` or `containsTextInName` or something to that effect, as there is a description field for api keys that is also text/string based.

PR #438, opened 2020-01-03
🐛 Allow refresh request to use expired jwt

anncatton:


src/test/java/bio/overture/ego/service/RefreshContextServiceTest.java
```diff
@@ -72,12 +72,11 @@ public void refresh_invalidContext_usedTokenIsDeleted() {
   @Test
   public void refresh_validContext_usedTokenIsDeleted() {
     val user1 = entityGenerator.setupUserWithRefreshToken("User One");
-    val user1Token = tokenService.generateUserToken(user1);
     val refreshToken1 = user1.getRefreshToken();
 
     assertTrue(refreshContextService.findById(refreshToken1.getId()).isPresent());
-    assertTrue(refreshContextService.getById(refreshToken1.getId()) != null);
-    refreshContextService.disassociateUserAndDelete(user1Token);
+    assertNotNull(refreshContextService.getById(refreshToken1.getId()));
```
andricDu: :+1:

src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -337,6 +333,15 @@ public Claims getTokenClaims(String token) {
     }
   }
 
+  public Claims getTokenClaimsIgnoreExpiry(String token) {
+    try {
+      return getTokenClaims(token);
+    } catch (ExpiredJwtException exception) {
+      log.info("Refreshing expired token!", exception);
```
andricDu: Let's avoid printing the stacktrace everytime this happens due to the `info` log level. Let's put the stack trace into `debug` while having the `info` print a single line. I've given an example here with some guards:
```java
  public Claims getTokenClaimsIgnoreExpiry(String token) {
    try {
      return getTokenClaims(token);
    } catch (ExpiredJwtException exception) {
      val claims = exception.getClaims();
      if (StringUtils.isEmpty(claims.getId())) {
        throw SomeSortofIllegalStateException
      }
      log.info("Refreshing expired token: {}", claims.getId());
      log.debug("Refreshing expired token!", exception);
      return claims;
    }
  }
```

PR #442, opened 2020-01-15
Mb/2.10.1 snapshot

andricDu:


pom.xml
```diff
@@ -5,7 +5,7 @@
 
     <groupId>bio.overture</groupId>
     <artifactId>ego</artifactId>
-    <version>2.9.1</version>
+    <version>2.10.1-SNAPSHOT</version>
```
lepsalex: 10.0-SNAPSHOT?

lepsalex: or actually 2.9.1-SNAPSHOT?

lepsalex: Dusan 1 : 0 Rob

PR #453, opened 2020-02-03
✨ 448 - add sorting to policy assoc

anncatton:
- adds search for user and group permissions.
- uses sub performant join when query and filters are empty, hibernate was breaking on `fetch().join()` setup


src/main/java/bio/overture/ego/repository/queryspecification/AbstractPermissionSpecification.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Stream;
+import javax.persistence.criteria.Predicate;
+import lombok.NonNull;
+import lombok.val;
+import org.springframework.data.jpa.domain.Specification;
+import org.springframework.security.core.parameters.P;
+
+public abstract class AbstractPermissionSpecification extends SpecificationBase {
```
rtisma: although this seems intuitive, you cannot refactor these Type bounded generics this way. <P> is not defined and so means nothing. Please replicate the logic before as was done in
https://github.com/overture-stack/ego/commit/5a5ce1e19847ba6204c06adbdf0947dd66cd0068#diff-29b6a77729ad77239fcf9b4a22d26536

src/main/java/bio/overture/ego/repository/queryspecification/AbstractPermissionSpecification.java
```diff
@@ -0,0 +1,74 @@
...
+      // TODO: EGO-452 This code results in redundant selects because it is not being fetched but
+      // instead only joined.
+      //  When using the joinFetch methodology, hibernate was throwing cryptic errors, so this is a
+      // sub performant fix until someone figures out the issue.
+      if (isNullOrEmpty(text) && filters.isEmpty()) {
+        val join = scb.leftJoin(Policy.class, policy);
+        return join.equalId(policyId);
+      }
+      val policySp = scb.leftJoinFetch(Policy.class, policy);
+      val permissionSp = scb.leftJoinFetch(AbstractPermission.class, OWNER);
```
rtisma: this shoulde be a left join on USer.class, not AbstractPermission.class. This code is joining AbstractPermission on AbstractPermission, which doesnt make sense. Here is how it was before
https://github.com/overture-stack/ego/blob/5a5ce1e198/src/main/java/bio/overture/ego/repository/queryspecification/UserPermissionSpecification.java#L57

src/main/java/bio/overture/ego/repository/queryspecification/UserPermissionSpecification.java
```diff
@@ -16,32 +16,7 @@
...
-  public static Specification<UserPermission> withUser(@NonNull UUID userId) {
-    return (root, query, builder) -> {
-      query.distinct(true);
-      Join<UserPermission, Policy> applicationJoin = root.join(OWNER);
-      return builder.equal(applicationJoin.<Integer>get(ID), userId);
-    };
-  }
-}
+@Slf4j
+public class UserPermissionSpecification extends AbstractPermissionSpecification {}
```
rtisma: bring back 
https://github.com/overture-stack/ego/blob/5a5ce1e198/src/main/java/bio/overture/ego/repository/queryspecification/UserPermissionSpecification.java

the owner field is of type User for UserPermissionSpecification and of type Group for GroupPermissionSpecification, so the join that is done in AbstractPermissionSpecification must define the owner type explicitly.

src/main/java/bio/overture/ego/service/AbstractPermissionService.java
```diff
@@ -318,7 +312,7 @@ private static AbstractPermission resolvePermissions(
     return permissions.get(0);
   }
 
-  private PolicyResponse convertToPolicyResponse(@NonNull P p) {
+  protected PolicyResponse convertToPolicyResponse(@NonNull P p) {
```
rtisma: make this static since there is no state involved

src/test/java/bio/overture/ego/controller/PolicyControllerTest.java
```diff
@@ -217,21 +219,206 @@ public void disassociatePermissionsFromUser_ExistingEntitiesAndRelationships_Suc
...
+
+    val responseStatus = response.getStatusCode();
+    assertEquals(responseStatus, OK);
+
+    val requestWithInvalidQuery =
+        initStringRequest()
+            .endpoint("/policies/%s/users", policyId)
+            .queryParam("limit", 20)
+            .queryParam("offset", 0)
+            .queryParam("query", "write");
```
rtisma: this will work since userPermissions are loaded, but if you tried querying by name (which lives in User), it would return 0 (or fail at the sql level), because of the issues mentioned above.

src/main/java/bio/overture/ego/controller/PolicyController.java
```diff
@@ -230,31 +222,109 @@ public void delete(
...
+  })
   @ApiResponses(
       value = {
         @ApiResponse(
             code = 200,
             message = "Get list of user ids with given policy id",
             response = String.class)
       })
-  public @ResponseBody List<PolicyResponse> findUserIds(
+  public @ResponseBody PageDTO<PolicyResponse> findUserIds(
```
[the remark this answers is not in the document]
anncatton: should I deprecate this and make a separate method for my changes?

src/main/java/bio/overture/ego/utils/IgnoreCaseSortDecorator.java
```diff
@@ -0,0 +1,71 @@
...
+  public Sort getSort() {
+    val orders =
+        delegate.getSort().stream()
+            .map(o -> new Sort.Order(o.getDirection(), getSortField(o.getProperty())).ignoreCase())
+            .collect(toUnmodifiableList());
+
+    return Sort.by(orders);
+  }
+
+  private String getSortField(String sort) {
```
anncatton: I wasn't sure about the best way to make this generic...i.e. right now this would break sorting on `name` on other entities

src/main/java/bio/overture/ego/controller/ApiKeyController.java
```diff
@@ -164,11 +165,11 @@ public ApiKeyController(
 
   @RequestMapping(method = DELETE, value = "/api_key")
   @ResponseStatus(value = OK)
-  public @ResponseBody String revokeApiKey(
+  public @ResponseBody GenericResponse revokeApiKey(
       @RequestHeader(value = "Authorization") final String authorization,
       @RequestParam(value = "apiKey") final String apiKey) {
     tokenService.revokeApiKey(apiKey);
-    return format("ApiKey '%s' is successfully revoked!", apiKey);
+    return createGenericResponse("ApiKey '%s' is successfully revoked!", apiKey);
```
rtisma: change `is` to `was`

src/main/java/bio/overture/ego/controller/PolicyController.java
```diff
@@ -190,23 +185,20 @@ public void delete(
...
   @ApiResponses(
       value = {@ApiResponse(code = 200, message = "Add user permission", response = String.class)})
-  public @ResponseBody String createUserPermission(
+  public @ResponseBody User createUserPermission(
       @RequestHeader(value = AUTHORIZATION, required = true) final String accessToken,
       @PathVariable(value = "id", required = true) UUID id,
       @PathVariable(value = "user_id", required = true) UUID userId,
       @RequestBody(required = true) MaskDTO maskDTO) {
-    userPermissionService.addPermissions(
+    return userPermissionService.addPermissions(
```
rtisma: :+1:

src/main/java/bio/overture/ego/controller/PolicyController.java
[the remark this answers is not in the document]
rtisma: since this is a breaking change, run 
`mvn versions:set` and change the version to `3.0.0-SNAPSHOT`

src/main/java/bio/overture/ego/repository/queryspecification/SpecificationBase.java
```diff
@@ -38,6 +39,27 @@
...
+      @NonNull CriteriaBuilder builder,
+      @NonNull Join join,
+      @NonNull String fieldName,
+      String fieldValue) {
+    val finalText = QueryUtils.prepareForQuery(fieldValue);
+
+    // Cast "as" String so that we can search ENUM types
+    return builder.like(builder.lower(join.get(fieldName).as(String.class)), finalText);
+  }
+
```
rtisma: delete these as they are unused

src/main/java/bio/overture/ego/service/AbstractPermissionService.java
rtisma: make this static

rtisma:
@anncatton i resolved all the stale/old comments. All the active comments here need to be addressed

PR #454, opened 2020-02-05
updated non root

henro001:
added new non root user


rtisma:
Main changes:
- when you run `API_HOST_PORT=8080 docker-compose up -d`, the `api` service is run a the `ego` user instead of `root` user
- the Dockerfile was updated to use an alpine jdk/jre version of java 11, since the previous one had security issues
- removed circleci stuff, since nothing has run in the last 7 months

rtisma:
closed since changes are in
https://github.com/overture-stack/ego/pull/455

PR #474, opened 2020-05-20
Bug missing json content type for several swagger ui endpoints#466

KevinFHartmann:


src/main/java/bio/overture/ego/controller/ApiKeyController.java
```diff
@@ -229,10 +234,10 @@ public ApiKeyController(
 
   /** DEPRECATED: GET /token to be removed in next major release */
   @Deprecated
+  @AdminScoped
```
andricDu, line 237: Is this a breaking change? Given that the first line of the method is a method call to `checkAdminOrOwner` emphasis on the word *owner*

andricDu: Yea if only ADMINs can list tokens, then this breaks the functionality in ARGO of users listing their own.

rtisma: GOOD EYE

src/main/java/bio/overture/ego/controller/ApiKeyController.java
```diff
@@ -85,42 +87,46 @@ public ApiKeyController(
...
   public @ResponseBody ApiKeyScopeResponse checkToken(
-      @RequestHeader(value = "Authorization") final String authToken,
+      @RequestHeader(value = "Authorization") final String authorization,
       @RequestParam(value = "token") final String token) {
 
-    return tokenService.checkApiKey(authToken, token);
+    return tokenService.checkApiKey(authorization, token);
   }
 
+  @AdminScoped
```
rtisma, line 118: this should be removed. the user should be able to get this info

PR #477, opened 2020-06-10
Rtisma 466 fixed browserstack

rtisma:


rtisma:
@andricDu The 4 spots that have this, are application scoped. They required Basic auith. So when you run swagger ui, that little "lock" symbol will by default add the header `Authorization: Bearer <jwt>`. By putting the `@RequestHearer` explicitly, its like a an override, where if the user put `Authorization: Basic <>`, then that header wil be used instead of the default global `Bearer` one

andricDu:
@rtisma but doesn't this change the functionality of the application? The error returned would be different if you don't include the Auth header.

andricDu:
To be precise, previously an exception would be thrown if you were missing the header. Now that exception is no longer thrown.

rtisma:
good catch. I just tested to make sure, if in secure mode and you dont provide the header, you get a 403, when in fact you should get a 400.Will make change now

andricDu:
To be honest, the correct error code should be 401 if no auth header is provided but that is outside the scope of this ticket.

My recommendation is that we just restore the existing functionality/behaviour and then we can create a ticket to eventually fix the 400s for missing header to be 401.

Issue #482, opened 2020-07-14
Migration for unique join rows cannot run if it already has duplicates


andricDu:
A migration will need to written that runs before the one that enforces unique constraints on the join tables. 

As the join tables do not have a single ID for each row, this will require a migration that does some juggling with intermediate temp tables. 

The migration will need to clear the duplicates from each join table.

andricDu:
https://github.com/overture-stack/ego/pull/483

andricDu:
3.1.1 tested in dev and qa

PR #483, opened 2020-07-14
migration to remove duplicate join table rows

andricDu:
We missed removing the duplicate rows for the add unique constraint migration. This is required for that one to run correctly.


andricDu:
https://github.com/overture-stack/ego/issues/482

PR #495, opened 2020-09-04
Rc/3.1.2

anncatton:


rtisma (changes requested):
This release include the new application permissions feature. This should be a minor version update and not a path. Please change the version to 3.2.0

andricDu (changes requested):
Would it be better to close this PR and recreate it with a branch rc/3.2.0?

PR #505, opened 2020-10-01
✨ 499 - Add sso filter for orcid

anncatton:


rtisma (changes requested):

src/main/java/bio/overture/ego/service/OrcidService.java
```diff
@@ -0,0 +1,57 @@
...
+        restTemplate.exchange(url, HttpMethod.GET, request, OrcidEmailResponse.class);
+
+    if (response.getStatusCode() == OK) {
+      val emails = response.getBody().getEmail();
+      val primaryEmail =
+          emails.stream()
+              .filter(OrcidEmail::isPrimary)
+              .filter(OrcidEmail::isVerified)
+              .findFirst()
+              .orElse(null)
```
rtisma, line 45: we need to handle the case that there is an empty set of emails. In the case that there are no emails, this stream would return null and then would call `getEmail()` on it, resulting in a NPE (null pointer exception), which would be misleading.

src/main/java/bio/overture/ego/service/OrcidService.java
```diff
@@ -0,0 +1,57 @@
...
+              .getEmail();
+      if (primaryEmail.isEmpty()) {
+        return Collections.singletonMap("error", "Could not fetch user details");
```
rtisma, line 48: static import. also, i see that spring internally looks for the key `error` in the transformMap method response, and throws an IvalidTokenException

src/main/java/bio/overture/ego/service/OrcidService.java
```diff
@@ -0,0 +1,57 @@
...
+      } else {
+        map.put("email", primaryEmail);
+        return map;
+      }
+    } else {
+      return Collections.singletonMap("error", "Could not fetch user details");
```
rtisma, line 54: static import

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -192,4 +200,23 @@ public FacebookFilter(OAuth2ClientResources client) {
...
+      super.setTokenServices(
+          new OAuth2UserInfoTokenServices(
+              client.getResource().getUserInfoUri(),
+              client.getClient().getClientId(),
+              super.restTemplate) {
+            @Override
+            protected Map<String, Object> transformMap(
+                Map<String, Object> map, String accessToken) {
+              val orcid = map.get("sub").toString();
+              OAuth2RestOperations restTemplate = getRestTemplate(accessToken);
```
rtisma, line 216: use `val`

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -192,4 +200,23 @@ public FacebookFilter(OAuth2ClientResources client) {
...
+      super("/oauth/login/orcid", client);
+      super.setTokenServices(
+          new OAuth2UserInfoTokenServices(
+              client.getResource().getUserInfoUri(),
+              client.getClient().getClientId(),
+              super.restTemplate) {
+            @Override
+            protected Map<String, Object> transformMap(
+                Map<String, Object> map, String accessToken) {
+              val orcid = map.get("sub").toString();
```
rtisma, line 215: this should be renamed to `subject`. Also, need to add error handling for when the subject is NOT orchid

rtisma (changes requested):

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
anncatton: the `map` returned here from the `userInfoUri` comes back in this structure:
`{id=https://orcid.org/{orcid}, sub={orcid}, name=null, family_name=string, given_name=string}`
Mentioned here: https://github.com/ORCID/ORCID-Source/blob/master/orcid-web/ORCID_AUTH_WITH_OPENID_CONNECT.md#authorization-code-flow
So the `sub` value will always be the orcid. Not sure if that's what you're referring to in `add error handling for when the subject is NOT orcid` ?

andricDu (approved):
Good stuff @anncatton

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
rtisma: Just being defensive. Add a simple check, and if it fails, return a 502 bad gateway error, indicating an invalid response from the upstream (orchid) server, since you were expecting "sub" to be "orcid" but was something else instead

rtisma (approved):

rtisma (approved):

andricDu:
Are we good for merge??? Just saw this is still in my feed.

andricDu:
Thanks @anncatton !!!!

PR #512, opened 2020-10-22
Provide Default Scope if none available

joneubank:
Also enforces Policy name characters to be {alphanumeric, _, -}

Required for 3.3.x to be deployed as this was blocking application authentication.

### DIFF
--- a/src/test/java/bio/overture/ego/controller/AbstractResolvablePermissionControllerTest.java
+++ b/src/test/java/bio/overture/ego/controller/AbstractResolvablePermissionControllerTest.java
@@ -27,9 +27,9 @@ public abstract class AbstractResolvablePermissionControllerTest<
     val group = getEntityGenerator().setupGroup(UUID.randomUUID().toString());
 
     // create policies
-    val readPolicy = getEntityGenerator().setupSinglePolicy("READ Policy");
-    val writePolicy = getEntityGenerator().setupSinglePolicy("WRITE Policy");
-    val denyPolicy = getEntityGenerator().setupSinglePolicy("DENY Policy");
+    val readPolicy = getEntityGenerator().setupSinglePolicy("READPolicy");
+    val writePolicy = getEntityGenerator().setupSinglePolicy("WRITEPolicy");
+    val denyPolicy = getEntityGenerator().setupSinglePolicy("DENYPolicy");
 
     // Add permissions to owner
     val r1 =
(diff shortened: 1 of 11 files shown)


src/test/java/bio/overture/ego/controller/AbstractResolvablePermissionControllerTest.java
```diff
@@ -97,9 +97,9 @@ public void resolveOwnerAndGroupPermissions_hasPermissionOverlap_Success() {
     val group3 = getEntityGenerator().setupGroup(UUID.randomUUID().toString());
 
     // create policies
-    val policy1 = getEntityGenerator().setupSinglePolicy("Policy 1");
-    val policy2 = getEntityGenerator().setupSinglePolicy("Policy 2");
-    val policy3 = getEntityGenerator().setupSinglePolicy("Policy 3");
+    val policy1 = getEntityGenerator().setupSinglePolicy("Policy1");
+    val policy2 = getEntityGenerator().setupSinglePolicy("Policy2");
```
andricDu, line 101: did you do this or a formatter? I don't care, just curious

src/test/java/bio/overture/ego/controller/TokensOnPermissionsChangeTest.java
```diff
@@ -175,7 +175,7 @@ public void downgradePermissionFromUser_ExistingToken_RevokeTokenSuccess() {
   @SneakyThrows
   public void denyPermissionFromUser_ExistingToken_RevokeTokenSuccess() {
     val user = entityGenerator.setupUser("UserFoo DenyPermission");
-    val policy = entityGenerator.setupSinglePolicy("song.abc");
+    val policy = entityGenerator.setupSinglePolicy("songabc");
```
andricDu, line 178: aw crap I forgot about this. SONG might use dots to delimit studies. Let me verify that we are not doing this in ICGC ARGO.

andricDu (approved):
Holding off on merge as I check how we are doing things in ARGO. Im pretty sure this is a breaking change for legacy icgc but that's no biggy.

src/test/java/bio/overture/ego/controller/AbstractResolvablePermissionControllerTest.java
andricDu: Oh wait I understand, no spaces allowed in policy name :+1:

andricDu:
So we don't use dots in our policy names in ARGO. So this should be fine.
@rtisma I just want to bring your attention to this change given how SONG used to expect scopes though I believe we have moved from that older way. 

@joneubank merge at your discretion.

rtisma:
Looking at song, there is no parsing of a scope. Song does a full string comparison against the scope
https://github.com/overture-stack/SONG/blob/d9757f5611fb9261da5214a90899045dc5630662/song-server/src/main/java/bio/overture/song/server/security/StudySecurity.java#L63
https://github.com/overture-stack/SONG/blob/d9757f5611fb9261da5214a90899045dc5630662/song-server/src/main/java/bio/overture/song/server/security/SystemSecurity.java#L53

Since we configure the prefix and suffix for scopes
https://github.com/overture-stack/SONG/blob/d9757f5611fb9261da5214a90899045dc5630662/song-server/src/main/resources/application.yml#L197-L201

adding dots in the policy name does not affect song.

joneubank:
@rtisma To be clear, we are REMOVING dots, not adding. Is there an issue with the policies currently being used there?

rtisma:
no there should not be an issue. 
if the policy is `X`, and the a permission is created for a group with mask `WRITE`, EGO will render the scope `X.WRITE`

in song-server config, i can config it to use the systemScope `X.WRITE`, but not the studyScope, since the study scope would look something like `X-MY_PROGRAM_ID.WRITE` 

so, song expects a delimiter in the context of a scope tied to a study, otherwise it does not expect a delimiter.

joneubank:
Thanks @rtisma @andricDu !

PR #517, opened 2020-10-30
remove aud claim from all token claims

jaserud:
Removes optional `aud` claims from TokenClaims. See spec for more details on "aud" claims https://tools.ietf.org/html/rfc7519#page-9

I removed the field entirely rather than return an empty array because from the spec it sounded to me like if the claim exists at all you have to identify a correct aud claim in the JWT consumer.

I also removed the claim from user too because it didn't make sense to keep that either. The aud claim was being set to all apps the user is connected with which seems redundant given scope names are for that. I can put it back if we do need it though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,11 @@ An example ego JWT is mentioned below:
     "iat": 1459458458,
     "exp": 1459487258,
     "jti": "56fd919accf2c1c7ec5c5d16",
+    "aud": [
+        "service1-id",
+        "service2-id",
+        "service3-id"
+    ],
     "context": {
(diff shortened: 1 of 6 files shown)


rtisma (approved):

andricDu (approved):

jaserud:
After giving it some thought I put back the aud claims in user tokens because those are being used actively. Their is a potential that some service expects this field in its parsing of a jwt and might bork if its not their.

andricDu:
That is a reasonable approach. LGTM!

PR #523, opened 2020-11-26
Remove user POST endpoint

anncatton:


src/test/java/bio/overture/ego/controller/CreateUserControllerTest.java
```diff
@@ -0,0 +1,761 @@
...
+    assertTrue(isNull(newUser.getEmail()));
+  }
+
+  @Test
+  public void createUser_BlankProviderId_BadRequest() {
+    val idToken = entityGenerator.setupUserIDToken(DEFAULT_PROVIDER_TYPE, "");
+
+    exceptionRule.expect(MalformedRequestException.class);
+    exceptionRule.expectMessage("Provider id cannot be blank.");
+    userService.getUserByToken(idToken);
```
rtisma, line 759: Leave a comment saying we dont need to test providerType being null, since we hardcode it to not be null via, SSOFilters

src/test/java/bio/overture/ego/service/DefaultProviderServiceTest.java
```diff
@@ -0,0 +1,50 @@
...
+  public void defaultProviderConfigured_Success() {
+    assertNotNull(configuredDefaultProvider);
+    defaultProviderService.findById(configuredDefaultProvider);
+    val defaultProvider = defaultProviderService.getById(configuredDefaultProvider);
+    assertNotNull(defaultProvider);
+    assertEquals(defaultProvider.getId(), configuredDefaultProvider);
+  }
+
+  @Test
+  public void defaultProviderConfigured_Failure() {
```
rtisma, line 41: just an fyi, to truly test the tripwire, we need to try and run spring, and midway update its context, so it can trigger the context refresh event and see the trip wire work

rtisma: you can try something like this
https://stackoverflow.com/a/55242370

rtisma (changes requested):
Great work, however some fixes are needed.

src/main/resources/flyway/conf/flyway.conf
```diff
@@ -134,6 +134,7 @@ flyway.locations = filesystem:src/main/resources/flyway/sql,classpath:db.migrati
 # Placeholders to replace in Sql migrations
 # flyway.placeholders.user=
 # flyway.placeholders.my_other_placeholder=
+flyway.placeholders.default_provider = GOOGLE
```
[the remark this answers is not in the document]
anncatton: i'm not sure if this is because our flyway implementation or just Flyway itself, but not having the value here causes an error when running  the `.fly` script: 
`No value provided for placeholder expressions: ${default_provider}.  Check your configuration!`

src/test/java/bio/overture/ego/controller/UserControllerTest.java
```diff
@@ -915,16 +551,16 @@ public void typeValidation_MalformedType_BadRequest() {
...
+    //    val templateR1 =
+    //        CreateUserRequest.builder()
+    //            .email(generateNonExistentName(userService) + "@xyz.com")
+    //            .status(APPROVED)
+    //            .preferredLanguage(ENGLISH)
+    //            .firstName("")
+    //            .lastName("")
+    //            .build();
+    //    val r1 = ((ObjectNode) MAPPER.valueToTree(templateR1)).put(TYPE, invalidType);
+    //    initStringRequest().endpoint("/users").body(r1).postAnd().assertBadRequest();
```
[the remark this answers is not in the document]
anncatton: I've added 3 tests for checking the provider info on an updateUser request. I'd like to keep other changes related to user updates in https://github.com/overture-stack/ego/issues/532, if that works? Just to separate some of these changes out a bit more.

src/test/java/bio/overture/ego/controller/UserControllerTest.java
```diff
@@ -947,16 +585,17 @@ public void preferredLanguageValidation_MalformedPreferredLanguage_BadRequest()
...
+    //        CreateUserRequest.builder()
+    //            .email(generateNonExistentName(userService) + "@xyz.com")
+    //            .status(APPROVED)
+    //            .type(USER)
+    //            .firstName("")
+    //            .lastName("")
+    //            .build();
+    //    val r1 = ((ObjectNode) MAPPER.valueToTree(templateR1)).put(PREFERREDLANGUAGE,
+    // invalidLanguage);
+    //    initStringRequest().endpoint("/users").body(r1).postAnd().assertBadRequest();
```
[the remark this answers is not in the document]
anncatton: One note about these malformed properties tests - at one point you mentioned removing them, but do you still want to do that? I had commented them out so i could reference them in https://github.com/overture-stack/ego/issues/532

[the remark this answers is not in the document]
anncatton: Disregard my comment about referencing the old createUser code - i deleted it. But i'm still thinking of moving any of these malformed request tests to an updateUserControllerTest in  #532. What do you think?

src/main/java/bio/overture/ego/config/DefaultProviderConfig.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+public class DefaultProviderConfig implements ApplicationListener<ContextRefreshedEvent> {
+
+  @Autowired private DefaultProviderService defaultProviderService;
```
rtisma, line 20: Lets avoid component injection this way. Create a constructor that can accept this dependency, and mark it final

src/main/java/bio/overture/ego/config/DefaultProviderConfig.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Value("${spring.flyway.placeholders.default_provider}")
+  private ProviderType configuredProvider;
```
rtisma, line 23: likewise with this. refer to previous comment

src/main/java/bio/overture/ego/config/DefaultProviderConfig.java
```diff
@@ -0,0 +1,48 @@
...
+import lombok.val;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+public class DefaultProviderConfig implements ApplicationListener<ContextRefreshedEvent> {
```
rtisma, line 18: This name is missleading. I would suggest something like `DefaultProviderGuardListener`

src/main/java/bio/overture/ego/config/DefaultProviderConfig.java
```diff
@@ -0,0 +1,48 @@
...
+    val storedProvider = storedProviderResult.get().getId();
+    checkState(
+        configuredProvider.equals(storedProvider),
+        "Configured default_provider '%s' does not match what was previously configured '%s'",
+        configuredProvider,
+        storedProvider);
+
+    log.info(
+        String.format(
+            "Configured default_provider '%s' matches previously configured '%s, finished initializing.",
```
rtisma, line 45: no initialization was done here actually. More like asserting the defined provider is consistent. Suggest changing to 

`...., finished bootstrap check`

src/main/java/bio/overture/ego/service/DefaultProviderService.java
```diff
@@ -0,0 +1,31 @@
...
+  private final DefaultProviderRepository defaultProviderRepository;
+
+  @Autowired
+  public DefaultProviderService(@NonNull DefaultProviderRepository defaultProviderRepository) {
+    super(DefaultProvider.class, defaultProviderRepository);
+    this.defaultProviderRepository = defaultProviderRepository;
+  }
+
+  @Override
+  public DefaultProvider getWithRelationships(ProviderType id) {
```
rtisma, line 26: add `@NonNull` to the parameter. It is good practice for public method parameters to have the `@NonNull` annottion, since the content of the incomming parameter is out of context of this class. For instance, the developer, cannot control how the public method is used outside this class. Private methods on the otherhand, are in the developer's control, and so the use of `@NonNull` is optional, if the scenario makes sense.

The point is, we want to catch null errors as early as possible in the callstack, so that debugging is easier and the fix can be done faster.

src/main/java/bio/overture/ego/service/UserService.java
```diff
@@ -174,11 +172,16 @@ public User getUserByToken(@NonNull IDToken idToken) {
         findByProviderTypeAndProviderId(providerType, providerId)
             .or(
                 () -> {
```
rtisma, line 174: This is a little hard to read. Refactor this into a method that is called in the `.or` statement.

rtisma: This comment is for the current functional component passed to the `.or()` statement

src/test/java/bio/overture/ego/service/DefaultProviderServiceTest.java
rtisma: @anncatton timebox this to a day. If not achievable, we will just validate in Dev/Qa

rtisma (changes requested):

src/main/java/bio/overture/ego/service/UserService.java
```diff
@@ -168,28 +166,10 @@ public User createFromIDToken(IDToken idToken) {
...
-                          log.info("User found, updating provider info.");
-                          foundUser.setProviderType(idToken.getProviderType());
-                          foundUser.setProviderId(idToken.getProviderId());
-                        });
-                    return userByEmailResult;
-                  }
-                  log.info("No email provided");
-                  return Optional.empty();
-                })
+            .or(() -> findByProviderTypeAndEmail(idToken))
```
rtisma, line 172: :+1:

rtisma (approved):

PR #524, opened 2020-11-26
Different durations for app and user jwts.

andricDu:
- User JWT duration is configured via `jwt.user.durationMs`
- App JWT duration is configured via `jwt.app.durationMs`
- Ran google code formatter
- Improved code quality and eliminated many warning from `TokenService.java`
- Removed deprecation annotation as non-deprecated method throws `NotImplemented`

### Did some: 
image

### DIFF
--- a/src/main/java/bio/overture/ego/service/TokenService.java
+++ b/src/main/java/bio/overture/ego/service/TokenService.java
@@ -109,11 +109,11 @@ public class TokenService extends AbstractNamedService<ApiKey, UUID> {
   private final UserRepository userRepository;
 
   /** Configuration */
-  private final int USER_JWT_DURATION;
+  private final int userJwtDuration;
 
-  private final int APP_JWT_DURATION;
+  private final int appJwtDuration;
 
-  private final int API_TOKEN_DURATION;
+  private final int apiTokenDuration;
 
   public TokenService(
       @NonNull TokenSigner tokenSigner,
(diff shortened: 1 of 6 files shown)


rosibaj (approved):

anncatton (approved):

src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -282,25 +281,33 @@ public String generateAppToken(Application application) {
...
+      decodedToken = Jwts.parser().setSigningKey(tokenKey).parseClaimsJws(token);
     } catch (JwtException e) {
       log.error("JWT token is invalid", e);
     }
     return (decodedToken != null);
   }
 
   public Jws<Claims> validateAndReturn(String token) {
-    Jws<Claims> decodedToken = null;
+    Jws<Claims> decodedToken;
```
rtisma, line 304: refactor so that `Optional<Jwa<Claims>> processValidToken(@NonNull String token)` does all the meat of the work, and then
```
public boolean isValidToken(String token){ 
   return processValidToken(token).isPresent(); 
}

public Jws<Claims> validateAndReturn(String token){
     return processValidToken(token).orElseThrow(() -> {
        log.error("JWT token was null when trying to validate and return.");
        return new ForbiddenException("Authorization is required for this action.");
     });
}
```

rtisma (changes requested):

src/main/java/bio/overture/ego/service/TokenService.java
andricDu: my man, that is legit

src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -99,17 +101,19 @@
...
   private final UserRepository userRepository;
 
   /** Configuration */
-  private int JWT_DURATION;
+  private final int USER_JWT_DURATION;
 
-  private int API_TOKEN_DURATION;
+  private final int APP_JWT_DURATION;
+
+  private final int API_TOKEN_DURATION;
```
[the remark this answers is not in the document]
andricDu: :+1:

rtisma (approved):

PR #535, opened 2020-12-16
530 remove name from user

anncatton:


rtisma (changes requested):
Great work, just some minor changes. In regards to the permission tests, that level of inheritance is not maintainable and generally discourage, however given the fact its the only way to create common test code (i.e common test code via composition is not supported), its acceptable for this special case.

src/main/java/bio/overture/ego/service/UserPermissionService.java
```diff
@@ -52,6 +52,13 @@ public UserPermissionService(
     this.apiKeyEventsPublisher = apiKeyEventsPublisher;
   }
 
+  protected PolicyResponse convertToPolicyResponse(@NonNull UserPermission userPermission) {
```
rtisma, line 55: add `@Override`

src/main/proto/Ego.proto
```diff
@@ -59,4 +58,6 @@ message User {
 
     google.protobuf.StringValue provider_type = 14;
     google.protobuf.StringValue provider_id = 15;
+
+    reserved 7;
```
rtisma, line 62: add `reserved name` and also add a comment 
```
// remove the name = 7 field as part of EGO-### ticket
```

rtisma (changes requested):

src/main/java/bio/overture/ego/service/UserPermissionService.java
anncatton: i removed the method from the AbstractPermissionService so there's nothing to override...AbstractNameablePermissionService has its own implementation that uses `owner.getName`

rtisma (approved):
Great work. Dont forget to update your "Summary" list of all the major changes  and breaking changes

PR #536, opened 2020-12-16
SameSite cookie fix. Allows session cookie to be configured.

andricDu:
## SameSite cookie setting for cross origin SSO
image

Allows the configuring of the `DefaultCookieSerializer` as part of `spring-session`.  This will allow us to set the `SameSite` attribute to `None` allowing for cross origin logins.

### Configuration 
With sensible defaults:
```yaml
login:
  nonce:
    name: LOGIN_NONCE
    sameSite: none
    secure: true
    maxAge: 120
    domainPattern: ^.+?\.(\w+\.[a-z]+)$
```

A `maxAge` of `-1` will tell the browser to delete the cookie on next close. 

### Usage
Should be transparent to the UI as these are `HttpOnly` cookies.

Example of new cookie in action:
```
Set-Cookie: LOGIN_NONCE=MWMxOTE2MWUtYjU5NS00ZGNkLWE5ZmItMDgzM2U4OTRkNDEy; Max-Age=120; Expires=Wed, 16 Dec 2020 12:55:43 -0500; Path=/; Secure; HttpOnly; SameSite=none
```

### Local Dev
- `SameSite: None` is only allowed when cookie is `Secure`. This means HTTPS only. Otherwise the browser will block the cookie form being set.
- If running Ego on local host, set `SameSite` to `lax` or `strict` and `secure` to `false`.

### DIFF
--- a/src/main/java/bio/overture/ego/config/SessionConfig.java
+++ b/src/main/java/bio/overture/ego/config/SessionConfig.java
@@ -18,6 +18,7 @@
 package bio.overture.ego.config;
 
 import java.util.concurrent.ConcurrentHashMap;
+import lombok.NonNull;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.session.MapSessionRepository;
@@ -30,7 +31,8 @@ public class SessionConfig {
 
   private final LoginNonceProperties properties;
 
-  public SessionConfig(@Autowired LoginNonceProperties properties) {
+  @Autowired
+  public SessionConfig(@NonNull LoginNonceProperties properties) {
     this.properties = properties;
   }
 
@@ -54,11 +56,11 @@ public class SessionConfig {
     serializer.setSameSite(properties.getSameSite());
     serializer.setUseSecureCookie(properties.isSecure());
     serializer.setCookieMaxAge(properties.getMaxAge());
+    serializer.setDomainNamePattern(properties.getDomainPattern());
 
     // These shouldn't be user configurable. ALWAYS make sure its HTTP Only.
     serializer.setUseHttpOnlyCookie(true);
     serializer.setCookiePath("/");
-    serializer.setDomainNamePattern("^.+?\\.(\\w+\\.[a-z]+)$");
     return serializer;
   }
 }
--- a/src/main/java/bio/overture/ego/config/SessionConfig.java
+++ b/src/main/java/bio/overture/ego/config/SessionConfig.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2020. The Ontario Institute for Cancer Research. All rights reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/bio/overture/ego/config/SessionConfig.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Use in memory data store. TODO: Add support for redis and psql data store configurable behind
+   * profiles.
+   *
+   * @return SessionRepository implementation.
+   */
+  @Bean
+  public MapSessionRepository sessionRepository() {
+    return new MapSessionRepository(new ConcurrentHashMap<>());
```
rtisma, line 45: :+1:

src/main/java/bio/overture/ego/config/SessionConfig.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.session.config.annotation.web.http.EnableSpringHttpSession;
+import org.springframework.session.web.http.CookieSerializer;
+import org.springframework.session.web.http.DefaultCookieSerializer;
+
+@EnableSpringHttpSession
+public class SessionConfig {
+
+  private final LoginNonceProperties properties;
+
+  public SessionConfig(@Autowired LoginNonceProperties properties) {
```
rtisma, line 33: add `@NonNull`

src/main/java/bio/overture/ego/config/SessionConfig.java
```diff
@@ -0,0 +1,64 @@
...
+    // These can be user configured
+    serializer.setCookieName(properties.getName());
+    serializer.setSameSite(properties.getSameSite());
+    serializer.setUseSecureCookie(properties.isSecure());
+    serializer.setCookieMaxAge(properties.getMaxAge());
+
+    // These shouldn't be user configurable. ALWAYS make sure its HTTP Only.
+    serializer.setUseHttpOnlyCookie(true);
+    serializer.setCookiePath("/");
+    serializer.setDomainNamePattern("^.+?\\.(\\w+\\.[a-z]+)$");
```
rtisma, line 61: @andricDu  Plugging this into regex101.com, `example.com` does not match, but `anything.example.com` does. Not an issue for us as domain will have more than 2 parts, but just making sure this is what we want

rtisma (changes requested):

src/main/java/bio/overture/ego/config/SessionConfig.java
andricDu: From the Spring Session docs: https://docs.spring.io/spring-session/docs/2.0.0.M2/reference/html5/guides/java-custom-cookie.html#custom-cookie-sample


> We customize the domain name pattern (a regular expression) to be ^.?\\.(\\w\\.[a-z]+)$ This allows sharing a session across domains and applications. If the regular expression does not match, no domain is set and the existing domain will be used. If the regular expression matches, the first grouping will be used as the domain. This means that a request to https://child.example.com will set the domain to example.com. However, a request to http://localhost:8080/ or http://192.168.1.100:8080/ will leave the cookie unset and thus still work in development without any changes necessary for production.

andricDu: Might as well make this configurable though. I can see someone locking down the allowed domains via a pattern.

andricDu: @rtisma I've moved it into the application.yml

rtisma (approved):

PR #539, opened 2020-12-18
Removes the now private spring-repos. Updates grpc.

andricDu:
image

- Removes the new private spring maven repos
- Updates gRPC dependency as the older version is not in maven central.


akachru-github (approved):
This indeed looks to change what you have described. 
LGTM.

andricDu:
> This indeed looks to change what you have described.
> LGTM.

I'm Commander Shepard and this is my favourite review on the Citadel.

PR #549, opened 2021-01-14
♻️  Rename providerId field to providerSubjectId

anncatton:


src/main/proto/Ego.proto
```diff
@@ -57,7 +57,7 @@ message User {
     repeated string scopes = 13;
 
     google.protobuf.StringValue provider_type = 14;
-    google.protobuf.StringValue provider_id = 15;
+    google.protobuf.StringValue provider_subject_id = 15;
```
rtisma, line 60: Just pointing out here, that protobuf changes are need special attention
https://stackoverflow.com/questions/45431685/protocol-buffer-does-changing-field-name-break-the-message

this is good, since its just a field name change.

rtisma (approved):

PR #553, opened 2021-01-19
rename default provider env var

anncatton:


rtisma (approved):

src/main/java/bio/overture/ego/config/DefaultProviderGuardListener.java
```diff
@@ -37,20 +36,17 @@ public void onApplicationEvent(ContextRefreshedEvent event) {
...
     checkState(
         storedProviderResult.isPresent(),
-        "Tripwire was not set! This means the flyway migration did not run yet.");
-    val storedProvider = storedProviderResult.get().getId();
-    checkState(
-        configuredProvider.equals(storedProvider),
-        "Configured default_provider '%s' does not match what was previously configured '%s'",
-        configuredProvider,
-        storedProvider);
+        "Configured default-provider '%s' does not match what was previously configured. Check the defaultprovidertripwire table.",
```
rtisma, line 43: it may be better, to output what was previously configured. this is more useful than "check the table". 
Configured default-provider '%s' does not match what was previously configured '%s)'`

rtisma (changes requested):

rtisma (approved):

PR #488, opened 2020-08-04
✨ 470 - Add resolved scopes to app jwt

anncatton:
Adds resolved scopes to app jwt.
Exposes `/oauth/token` endpoint in swagger

PR #560, opened 2021-02-08
implement first user is admin option

blabadi:
add tests to cover it


blabadi:
tests are failing because some data is not cleaned up correctly after previous tests .. looking into why

src/test/java/bio/overture/ego/service/FirstUserAsAdminTest.java
```diff
@@ -0,0 +1,43 @@
...
+@ActiveProfiles("test")
+@TestPropertySource(properties = "default.user.firstUserAsAdmin=true")
+@Transactional
+public class FirstUserAsAdminTest {
+
+  @Autowired private UserService userService;
+  @Autowired private EntityGenerator entityGenerator;
+
+  @Test
+  public void testOnlyFirstUserShouldBeAdminByDefault() {
```
anncatton, line 30: Can we convert these tests to controller tests? Rob has indicated that these are more reliable when it comes to hibernate weirdness: https://github.com/overture-stack/ego/pull/488#pullrequestreview-463423170 
Maybe in the `CreateUserControllerTest`, to take advantage of the idtoken mock?

src/test/java/bio/overture/ego/service/UserServiceTest.java
```diff
@@ -28,6 +30,16 @@
   @Autowired private RefreshContextService refreshContextService;
   @Autowired private TokenService tokenService;
 
+  @Test
+  public void testFirstUserShouldNotBeAdminByDefault() {
+    userService.getRepository().deleteAll();
```
anncatton, line 35: Same as above comment

src/test/java/bio/overture/ego/service/FirstUserAsAdminTest.java
blabadi: I'll try to add another test there, but I don't want to spend more time than needed for this ticket, it's too small anyway. 

unless there is a clear reason why this is not sufficient to catch regressions I don't think we should do controller level tests for everything, they come at a cost

src/test/java/bio/overture/ego/utils/EntityGenerator.java
```diff
@@ -154,8 +154,9 @@ public User setupUser(
         user -> setupUser(user, USER, UUID.randomUUID().toString(), DEFAULT_PROVIDER_TYPE));
   }
 
-  public void setupTestUsers() {
-    setupUsers("First User", "Second User", "Third User");
+  public List<User> setupTestUsers() {
+    val users = setupUsers("First User", "Second User", "Third User");
```
blabadi, line 158: I don't know why I didn't return directly

blabadi:
just noticed the test failure.. looking into it

andricDu (approved):

src/test/java/bio/overture/ego/service/FirstUserAsAdminTest.java
blabadi: Adding controller test opened a can of worms because using @TestPropertySource breaks spring context between tests:

so TestSourceProperty is problematic for the controller tests not sure how that collides, but spring always does nasty shit when it comes to tests and cache context and what not, so based on the time I already wasted on this I'll remove the controller test

PR #562, opened 2021-02-17
🐛 561 - Modify update user logic

anncatton:


blabadi (approved):

src/main/java/bio/overture/ego/model/dto/UpdateUserRequest.java
```diff
@@ -31,6 +32,8 @@
   private UserType type;
   private StatusType status;
   private LanguageType preferredLanguage;
-  @NonNull private ProviderType providerType;
-  @NonNull private String providerSubjectId;
+  private String firstName;
+  private String lastName;
+  @NotNull private ProviderType providerType;
```
andricDu, line 37: Do we want to be able to change provider information via REST API?

andricDu (changes requested):

src/main/java/bio/overture/ego/model/dto/UpdateUserRequest.java
anncatton: I included the provider info here to add an extra validation on the user's provider sub id & type when updating:
https://github.com/overture-stack/ego/blob/8bd699b784bb442458239f150cb073450e3f1521/src/main/java/bio/overture/ego/service/UserService.java#L511
It was my intention to not allow the provider info to be changed, the request would error out if the incoming user provider info didn't match our stored user. But maybe I missed something?

src/main/java/bio/overture/ego/controller/UserController.java
```diff
@@ -146,7 +146,7 @@ public UserController(
   }
 
   @AdminScoped
-  @RequestMapping(method = RequestMethod.PUT, value = "/{id}")
+  @RequestMapping(method = RequestMethod.PATCH, value = "/{id}")
```
andricDu, line 149: :fireworks:

andricDu (approved):
Thanks!

PR #572, opened 2021-03-08
🐛 547 - Fix scope for linkedin filter

anncatton:
Adds explicit scopes for LinkedIn
Adds handling for new api responses

### DIFF
--- a/src/main/java/bio/overture/ego/model/dto/LinkedinEmail.java
+++ b/src/main/java/bio/overture/ego/model/dto/LinkedinEmail.java
@@ -1,0 +1,36 @@
+package bio.overture.ego.model.dto;
+
+import static java.util.Objects.isNull;
+
+import bio.overture.ego.model.enums.LinkedinContactType;
+import bio.overture.ego.view.Views;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonView;
+import java.util.LinkedHashMap;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@JsonInclude
+@JsonView(Views.REST.class)
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class LinkedinEmail {
+
+  @JsonProperty("handle~")
+  private LinkedHashMap<String, String> handle;
+
+  private boolean primary;
+  private LinkedinContactType type;
+
+  public String getEmail() {
+    if (isNull(this.handle) || isNull(this.handle.get("emailAddress"))) {
+      return null;
+    }
+    return this.handle.get("emailAddress");
+  }
+}
--- a/src/main/java/bio/overture/ego/model/dto/LinkedinEmail.java
+++ b/src/main/java/bio/overture/ego/model/dto/LinkedinEmail.java
@@ -1,7 +1,5 @@
 package bio.overture.ego.model.dto;
 
-import static java.util.Objects.isNull;
-
 import bio.overture.ego.model.enums.LinkedinContactType;
 import bio.overture.ego.view.Views;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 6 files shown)


src/main/java/bio/overture/ego/model/dto/LinkedinEmail.java
```diff
@@ -0,0 +1,36 @@
...
+public class LinkedinEmail {
+
+  @JsonProperty("handle~")
+  private LinkedHashMap<String, String> handle;
+
+  private boolean primary;
+  private LinkedinContactType type;
+
+  public String getEmail() {
+    if (isNull(this.handle) || isNull(this.handle.get("emailAddress"))) {
```
joneubank, line 31: Maybe you want to consider having a default value for `handle` as an empty Map so that this check isn't needed. Lombok has `@Default.Builder` annotation to handle this on the handle property, but some stack overflow is showing that it is buggy. Might be worth trying instead of having to check for nulls on every Map created via the Lombok constructors.

src/main/resources/application.yml
```diff
@@ -108,8 +108,10 @@ linkedIn:
     accessTokenUri: https://www.linkedin.com/oauth/v2/accessToken
     userAuthorizationUri: https://www.linkedin.com/oauth/v2/authorization
     clientAuthenticationScheme: query
+    scope: "r_liteprofile r_emailaddress"
   resource:
-    userInfoUri: https://api.linkedin.com/v1/people/~:(email-address,first-name,last-name)?format=json
+    userInfoUri: https://api.linkedin.com/v2/me
+    userEmailUri: https://api.linkedin.com/v2/clientAwareMemberHandles?q=members&projection=(elements*(primary,type,handle~))
```
joneubank, line 114: are `userInfoUri` and `userEmailUri` really configuration variables that need to be in our application.yml? we probably just want a `apiHost: https://api.linkedin.com/v2` and then the paths to those URIs can be built in application logic - no person spinning up Ego would want to customize or configure those paths, and would actually break the application if they did.

src/main/java/bio/overture/ego/service/LinkedinService.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Autowired
+  public LinkedinService(@NonNull @Value("${linkedIn.resource.userEmailUri}") String userEmailUri) {
+    this.userEmailUri = userEmailUri;
+  }
+
+  public Map<String, Object> getPrimaryEmail(
+      OAuth2RestOperations restTemplate, @NonNull Map<String, Object> map) {
+    val headers = new HttpHeaders();
+    headers.set("Accept", "application/json");
```
joneubank, line 38: Spring constant for `"Accept"`: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html#ACCEPT

joneubank: Spring constant for `"application/json"`: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html#APPLICATION_JSON

src/main/java/bio/overture/ego/service/LinkedinService.java
```diff
@@ -0,0 +1,76 @@
...
+        return singletonMap("error", "A primary email was not found for the user.");
+      } else {
+        val email = primaryEmail.get();
+
+        map.put("email", email);
+        map.put("given_name", map.get("localizedFirstName"));
+        map.put("family_name", map.get("localizedLastName"));
+        return map;
+      }
+    } else {
```
joneubank, line 71: If you want to simplify sections like this with lots of if/else nesting, you can invert the if statements to check for the error case and do your throw/return work first, and then continue without an else block if there are no errors or issues:

```
// Invert the if condition to check for a failure instead of check for success
if( !(response.getStatusCode() == OK && response.hasBody() && response.getBody() != null)) {
    log.error("Response body was empty.");
    return singletonMap("error", "Could not fetch user details");
}

val emails = response.getBody().getElements();
val primaryEmail =
    emails.stream()
        .filter(e -> e.getType() == LinkedinContactType.EMAIL)
        .filter(LinkedinEmail::isPrimary)
        .findFirst()
        .map(LinkedinEmail::getEmail);

// You already used phrased this if statement to check for error
if (primaryEmail.isEmpty()) {
    return singletonMap("error", "A primary email was not found for the user.");
}
// Don't need the else block

val email = primaryEmail.get();

map.put("email", email);
map.put("given_name", map.get("localizedFirstName"));
map.put("family_name", map.get("localizedLastName"));
return map;
```

This is just an alternative to reduce all the nesting/indentation in a method.

joneubank (changes requested):
Bunch of notes, but the only changes we require 100% are the config changes I mentioned. If the URI path won't be configurable then it should be part of the application and not in the config file.

src/main/java/bio/overture/ego/service/LinkedinService.java
anncatton: nice! will do

src/main/java/bio/overture/ego/model/dto/LinkedinEmail.java
joneubank: @anncatton did you test this and did it work?  It looks like the correct approach but I saw numerous SO tickets about it not working.

joneubank (approved):

PR #579, opened 2021-03-21
🚸 569 - SSO error handling

anncatton:
Adds error handling for providers that allow empty/non-visible email fields (tested with Orcid)
Adds error handling for a user denying ego access to their provider info
Adds `errorRedirectUri` field to ego applications so login errors can be handled on the client side.


src/main/java/bio/overture/ego/model/enums/JavaFields.java
```diff
@@ -65,4 +65,5 @@
   public static final String REFRESH_ID = "refreshId";
   public static final String PROVIDERTYPE = "providerType";
   public static final String PROVIDERSUBJECTID = "providerSubjectId";
+  public static final String ERRORREDIRECTURI = "errorRedirectUri";
```
joneubank, line 68: Makes me want to refactor our naming convention here to use underscores cause I can't read this. `REFRESH_ID` uses underscores so maybe we can also use `ERROR_REDIRECT_URI`. Bonus points if you want to also update `PROVIDER_SUBJECT_ID` and others.

joneubank:
@anncatton Can we get some PR description and some ticket linkage here?

anncatton:
> @anncatton Can we get some PR description and some ticket linkage here?

i'll add some details. ticket is linked tho - is it not showing?

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -61,6 +69,53 @@ public void onAuthenticationSuccess(
   private OrcidService orcidService;
   private GithubService githubService;
 
+  private AuthenticationFailureHandler failureHandler =
+      (request, response, exception) -> {
+        val rootExceptionThrowable = exception.getCause();
+        val application =
+            applicationService.getByClientId(
+                (String) request.getSession().getAttribute("ego_client_id"));
```
joneubank, line 77: Is it possible that we have a user not an application attempting to sign in and get caught up in this error handler? What happens in all the application and redirect specific code here when its a user with no client id or redirect uri?

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -61,6 +69,53 @@ public void onAuthenticationSuccess(
...
+            } catch (IllegalArgumentException e) {
+              log.warn(format("Invalid provider: '%s'", provider));
+            }
+            response.setStatus(403);
+            response.sendRedirect(errorUri.build().toString());
+          } catch (URISyntaxException e) {
+            log.error("Invalid redirect uri");
+          }
+        } else if (rootExceptionThrowable instanceof UserDeniedAuthorizationException) {
+          // User denies Ego access/cancels login
```
joneubank, line 103: Thank you for this comment, I thought the exception implied that the user was denied, not that the user did the denying. Maybe the name is not clear enough. `DeniedByUserException` or something?

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -61,6 +69,53 @@ public void onAuthenticationSuccess(
...
+          try {
+            errorUri = new URIBuilder(errorRedirect);
+            errorUri.addParameter("error_code", "403");
+            // setting explicitly to "access_denied" because LinkedIn error_code differs from other providers
```
joneubank, line 107: This is also a good comment but could use more detail about what each provider expects. Do we need seperate responses for linked in vs. other providers?

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -93,6 +148,7 @@ public OAuth2SsoChildFilter(String path, OAuth2ClientResources client) {
       OAuth2RestTemplate template = new OAuth2RestTemplate(client.getClient(), oauth2ClientContext);
       super.setRestTemplate(template);
       super.setAuthenticationSuccessHandler(simpleUrlAuthenticationSuccessHandler);
+      super.setAuthenticationFailureHandler(failureHandler);
```
joneubank, line 151: Spring Magic!

src/main/java/bio/overture/ego/security/OAuth2UserInfoTokenServices.java
```diff
@@ -60,7 +62,7 @@ public IDToken extractPrincipal(Map<String, Object> map) {
     if (map.get("email") instanceof String) {
       email = (String) map.get("email");
     } else {
-      return null;
+      throw new NoPrimaryEmailException("no_primary_email");
```
joneubank, line 65: This message is not useful, we already know there was no primary email from the error type. Maybe there are no relevant details to add but could still be proper readable language instead of this snake_case token.

joneubank (commented):
Couple questions. Looks good overall. Maybe @andricDu also wants to review?

src/main/java/bio/overture/ego/model/enums/JavaFields.java
andricDu: @joneubank is right. Constants should use snake case upper caps.

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
andricDu: Nothing magical about an aptly named method :smiley:

src/main/java/bio/overture/ego/model/enums/JavaFields.java
anncatton: just in the JavaFields tho, right? (not sql fields...not looking to rename a bunch of columns atm :) )

joneubank: just the java variables. not the sql column names.

joneubank:
> i'll add some details. ticket is linked tho - is it not showing?

It is showing now, thanks @anncatton

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
anncatton: this is a built in exception from Spring :)

joneubank: Alrighty. GOod comment then.

src/main/java/bio/overture/ego/security/OAuth2UserInfoTokenServices.java
anncatton: Yeah on my first go at this i'm just passing the message as a param to client side, but I can set that where the exception gets caught, and make this message more readable instead.

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -61,6 +69,53 @@ public void onAuthenticationSuccess(
...
+                (String) request.getSession().getAttribute("ego_client_id"));
+        String errorRedirectUri =
+            (String) request.getSession().getAttribute("ego_error_redirect_uri");
+
+        val errorRedirect = Redirects.getErrorRedirectUri(application, errorRedirectUri);
+        URIBuilder errorUri;
+
+        if (rootExceptionThrowable instanceof NoPrimaryEmailException) {
+          try {
+            val reqUri = new ArrayList<String>(asList(request.getRequestURI().split("/")));
```
andricDu, line 86: `new ArrayList<>(...)`

The generic can be inferred.

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
```diff
@@ -61,6 +69,53 @@ public void onAuthenticationSuccess(
...
+            response.setStatus(403);
+            response.sendRedirect(errorUri.build().toString());
+          } catch (URISyntaxException e) {
+            log.error("Invalid redirect uri");
+          }
+        } else if (rootExceptionThrowable instanceof UserDeniedAuthorizationException) {
+          // User denies Ego access/cancels login
+          try {
+            errorUri = new URIBuilder(errorRedirect);
+            errorUri.addParameter("error_code", "403");
```
andricDu, line 106: `error_code` and `error_type` look like good candidates for constants.

andricDu (changes requested):

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
anncatton: Just tested this with linkedin now that it's working, looks like they come back with an `OAuth2Exception` with an `error="invalid_request"` when the user denies access...so i will need to handle that separately.

anncatton: Actually I can handle both those errors with this block. It might be useful to check the `error_code` too to make sure I'm handling the correct OAuth2Exception scenario

anncatton: Ooh even better: `Condition 'rootExceptionThrowable instanceof UserDeniedAuthorizationException' covered by subsequent condition 'rootExceptionThrowable instanceof OAuth2Exception'`

src/main/java/bio/overture/ego/security/OAuth2SsoFilter.java
joneubank: Ignore this comment, I forgot that User logins occur through applications so they all require an application client id in the request. No issues here.

joneubank (approved):
Looks like its all been addressed. Thanks for the work here.

andricDu (approved):

Issue #547, opened 2021-01-14
Linkedin login is broken


rtisma:
After creating a linkedin app and configuring ego to use the clientId and clientSecrets, when you try to login with linkedin, an error occurs and you are redirected to the url
```
https://<ego-domain>/api/oauth/login/linkedin?error=invalid_request&error_description=You+need+to+pass+the+%22scope%22+parameter&state=vKgzXM
```

as the url suggests, the `scope` parameter needs to be passed.

anncatton:
linkedin got rid of default scopes a while back apparently https://docs.microsoft.com/en-us/linkedin/shared/references/migrations/default-scopes-migration

b-f-chan:
Need to modify our OAuth flow to explicitly request the scopes & map/identify them per the steps in the migration link above

b-f-chan:
Removed enhancement, we should fix this because LinkedIn is a supported provider, and this is also blocking our testing for this provider

anncatton:
@joneubank suggested in the pr review for this ticket moving the `userInfoUri` and the `userEmailUri` out of the application properties as these are basically static values, but since all of the IdPs are currently set up this way, maybe this can be introduced later as a refactor

b-f-chan:
Tested on staging in EGO UI and DMS UI, LinkedIn login works

Can close

b-f-chan:
`userInfoUri` and `userEmailUri` to be moved out of application properties in separate ticket, ego#584

Closing ticket

PR #580, opened 2021-03-22
🔖 Rc/4.2.0

anncatton:
#547


joneubank (approved):
This SHOULD be a patch release not a minor version release since it is a bug fix and no features are added. However, there were changes to the configuration file (application.yml) and the functionality will not work without those updates, so this doesn't qualify as a patch.

Approving as a minor version increase because of the configuration file changes.

PR #595, opened 2021-04-13
use chart 3.0.0

blabadi:
Ego is ready to use chart 3.0.0, already deployed in the env and this PR is for future CDs


anncatton (approved):
🎉

PR #647, opened 2022-05-23
Remove Auth from API Key Check

joneubank:
Removed authorization check from `TokenService.checkApiKey` . Auth checking is done at the request level, it should not be repeated here. See attached ticket for full details.

### DIFF
--- a/src/main/java/bio/overture/ego/model/dto/ApiKeyScopeResponse.java
+++ b/src/main/java/bio/overture/ego/model/dto/ApiKeyScopeResponse.java
@@ -12,7 +12,6 @@ import lombok.Getter;
 @JsonView(Views.REST.class)
 public class ApiKeyScopeResponse {
   private UUID user_id;
-  private String client_id;
   private Long exp;
   private Set<String> scope;
 }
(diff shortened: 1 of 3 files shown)


src/main/java/bio/overture/ego/model/dto/ApiKeyScopeResponse.java
```diff
@@ -12,7 +12,6 @@
 @JsonView(Views.REST.class)
 public class ApiKeyScopeResponse {
   private UUID user_id;
-  private String client_id;
```
blabadi, line 15: not sure if removing this will break song/score.
since you still have the authorization is there a way to keep this if needed ?

joneubank: Yes, its doable with a bit of moving around where the response object is built and pulling the client ID from the auth.

The `client_id` value is just the application client_id that made the request. No application should need to see this when they generated the original request anyways... so I removed it because it seems to have no value and complicated building the response.

I would prefer removing any dependency on this field vs adding it back in, unless you can see a reason to provide this.

blabadi: yeah I agree that it's better to remove it I'm just wondering if it will break anything, I checked that score is using spring class (RemoteTokensService) which maybe expecting this for some reason, because it seems to be here intentionally, I just don't want to break other services, if you are sure of it no problem from my end

joneubank: Its a valid concern. Hold off on this review until I can confirm what the spring RemoteTokensService expects in the endpoint.

blabadi (approved):

PR #682, opened 2023-04-03
check_api_key should not return HTTP401

Azher2Ali:
In order to return messages a boolean property "valid".
When the apikey is invalid, return HTTP 200 with response body like:

{
  "valid": false,
  "error": "...message explaining why invalid"
}

 **_Expected Output_** 
Screenshot 2023-04-03 at 12 34 19 PM

### DIFF
--- a/src/main/java/bio/overture/ego/service/TokenService.java
+++ b/src/main/java/bio/overture/ego/service/TokenService.java
@@ -399,16 +399,19 @@ public class TokenService extends AbstractNamedService<ApiKey, UUID> {
   public ApiKeyScopeResponse checkApiKey(final String apiKey) {
     if (apiKey == null) {
       log.debug("Null apiKey");
-      throw new InvalidTokenException("No apiKey field found in POST request");
+      throw new ClientInvalidTokenException("No apiKey field found in POST request");
     }
 
     log.debug(format("apiKey ='%s'", apiKey));
-
     val aK =
-        findByApiKeyString(apiKey).orElseThrow(() -> new InvalidTokenException("ApiKey not found"));
+        findByApiKeyString(apiKey)
+            .orElseThrow(
+                () -> {
+                  return new ClientInvalidTokenException("ApiKey not found");
+                });
 
     if (aK.isRevoked())
-      throw new InvalidTokenException(
+      throw new ClientInvalidTokenException(
           format("ApiKey \"%s\" has expired or is no longer valid. ", apiKey));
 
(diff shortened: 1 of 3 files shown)


src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -399,16 +399,19 @@ private String getSignedToken(TokenClaims claims) {
   public ApiKeyScopeResponse checkApiKey(final String apiKey) {
     if (apiKey == null) {
       log.debug("Null apiKey");
-      throw new InvalidTokenException("No apiKey field found in POST request");
+      throw new ClientInvalidTokenException("No apiKey field found in POST request");
```
joneubank: no key provided should be a 400 for invalid request

src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -399,16 +399,19 @@ private String getSignedToken(TokenClaims claims) {
...
     }
 
     log.debug(format("apiKey ='%s'", apiKey));
-
     val aK =
-        findByApiKeyString(apiKey).orElseThrow(() -> new InvalidTokenException("ApiKey not found"));
+        findByApiKeyString(apiKey)
+            .orElseThrow(
+                () -> {
+                  return new ClientInvalidTokenException("ApiKey not found");
```
joneubank: apikey not found should be a 400 for invalid request

src/main/java/bio/overture/ego/service/TokenService.java
```diff
@@ -399,16 +399,19 @@ private String getSignedToken(TokenClaims claims) {
...
+                });
 
     if (aK.isRevoked())
-      throw new InvalidTokenException(
+      throw new ClientInvalidTokenException(
```
joneubank, line 414: 👍

joneubank (changes requested):
Need to properly respond when the request is invalid from the user (missing apikey)

src/main/java/bio/overture/ego/service/TokenService.java
Azher2Ali: 400 means, do you refer to invalid token exception?

joneubank: `InvalidRequestException` I think?

Azher2Ali (commented):
As per the request , the changes are made

joneubank (approved):
Thanks!

PR #683, opened 2023-04-05
Changes made and updated the HTTP response as per the TEST cases results

Azher2Ali:
Changes made and updated the HTTP response as per the TEST cases results

### DIFF
--- a/src/test/java/bio/overture/ego/controller/ApiKeyControllerTest.java
+++ b/src/test/java/bio/overture/ego/controller/ApiKeyControllerTest.java
@@ -323,7 +323,7 @@ public class ApiKeyControllerTest extends AbstractControllerTest {
     val response = initStringRequest().endpoint("o/check_api_key").body(params).post();
 
     val statusCode = response.getStatusCode();
-    assertEquals(HttpStatus.UNAUTHORIZED, statusCode);
+    assertEquals(HttpStatus.OK, statusCode);
   }
 
   @SneakyThrows
@@ -358,7 +358,7 @@ public class ApiKeyControllerTest extends AbstractControllerTest {
     val response = initStringRequest().endpoint("o/check_api_key").body(params).post();
 
     val statusCode = response.getStatusCode();
-    assertEquals(HttpStatus.UNAUTHORIZED, statusCode);
+    assertEquals(HttpStatus.BAD_REQUEST, statusCode);
   }
 
   @SneakyThrows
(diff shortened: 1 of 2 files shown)


joneubank (changes requested):
Thanks for updating the unit tests as well...

For the test cases where the API Key is rejected, its good that we are responding with `OK` but we should also be checking that the response includes `valid: false`

src/test/java/bio/overture/ego/controller/ApiKeyControllerTest.java
```diff
@@ -358,7 +358,10 @@ public void checkInvalidApiKey() {
     val response = initStringRequest().endpoint("o/check_api_key").body(params).post();
 
     val statusCode = response.getStatusCode();
-    assertEquals(HttpStatus.UNAUTHORIZED, statusCode);
+    /*assertEquals(HttpStatus.BAD_REQUEST, statusCode);
+    assertEquals("false", response.getStatusCode());*/
```
joneubank, line 362: Please remove unused comments.

src/test/java/bio/overture/ego/controller/ApiKeyControllerTest.java
```diff
@@ -358,7 +358,10 @@ public void checkInvalidApiKey() {
...
+      
+      assertEquals(HttpStatus.BAD_REQUEST, "false");*/
```
joneubank, line 364: We should be doing tests on the structure we expect to see:

statusCode = OK
body.valid = false

src/test/java/bio/overture/ego/controller/TokensOnPermissionsChangeTest.java
```diff
@@ -102,7 +102,7 @@ public void deletePermissionFromUser_ExistingToken_RevokeSuccess() {
         initStringRequest(tokenHeaders).endpoint("/o/check_api_key?apiKey=%s", apiKey).post();
 
     // Should be revoked
-    assertEquals(checkTokenAfterDeleteResponse.getStatusCode(), HttpStatus.UNAUTHORIZED);
+    assertEquals(checkTokenAfterDeleteResponse.getStatusCode(), HttpStatus.OK);
```
joneubank, line 105: all the checks for revoked token in this test file are incomplete. a response with status OK can be either valid or invalid. We also need to check if the response has the property `valid: false`

joneubank (changes requested):
This needs attention. The code does not compile as written and does not test for the case we expect. I left comments with additional details.

joneubank:
I've been doing some thinking about these changes... I think we should remove the changes from the other PR instead of pushing through with test case updates. The changes this causes will affect consuming applications.. this is a breaking change that I didn't prepare for thoroughly enough.

If we are going to make these changes, we need to provide a new endpoint (perhaps versioned as v2) and have clearer documentation about the response type. We should also include the same "valid" property with `valid: true` on success.  All in all, this was poorly conceived and prepared for by myself and for now I think we should revert... I'll help prep a PR to pull these changes out.

justincorrigible:
Closing as result of https://github.com/overture-stack/ego/pull/684

Issue #680, opened 2023-03-27
Create Admin CRUD Endpoints for Visa - Permission Mapping


Buwujiu:
## Detailed Description
Implement endpoints:
1. Get mappings by visa Id
2. Get mappings by policy Id, just the policy, not the mask
3. 1 endpoint to update or create visa permissions, 1 visa - many permissions. 
  - each visa should only have 1 entry per policy, example: SONG. READ, if the request is to update to SONG.WRITE, then update it to be SONG.WRITE, do not keep both. 
4. 1 endpoint to remove mappings by VISA id AND policy name(or ID? as long as it's consistent with the rest of the API). 1 visa - many policies.

All endpoints:
- Only accessible to admins.
- should have pagination 
- should have basic error handling

Issue #679, opened 2023-03-27
Add Visa - Permission Mapping to Ego DB


Buwujiu:
## Detailed Description
Add many to many relationship to Visa - Permissions.
A permission is a policy (ACL entity) and scope(mask) in ego. Example: `SONG.READ`.

## Possible Implementation
Refer to the note in https://wiki.oicr.on.ca/display/OV/Investigation+-+Ga4gh+passports+in+Ego#InvestigationGa4ghpassportsinEgo-VisaTypes

Possible table structure: 
```
CREATE TABLE ACLVISAPERMISSION (
  id                      UUID PRIMARY KEY,
  entity                  UUID,
  visaId                  UUID,
  mask                    ACLMASK NOT NULL,
  FOREIGN KEY (entity)    REFERENCES ACLENTITY(id),
  FOREIGN KEY (sid)       REFERENCES GA4GHVISA(id)
);
 
```

PR #688, opened 2023-04-25
committing changes related to Visa permissions schema

Azher2Ali:
Added many to many relationship to Visa - Permissions.
Also created a service class which does not impact #679 and which is further used in #680


joneubank (approved):
Everything looks good here.

I wonder if we should rename `entity` to `policyId` since thats what it is referencing. Took me a while to determine which property was referencing the ID from ACL policies since `entity` is so generic.

Azher2Ali:
> wonder if we should rename `entity` to `policyId` since thats what it is referencing. Took me a while to determine which property was referencing the ID from ACL policies since `entity` is so generic.

This change has been taken as part of current ticket #680

PR #690, opened 2023-04-26
Feature/visa permission api

### LINKED ISSUE #680: Create Admin CRUD Endpoints for Visa - Permission Mapping

Azher2Ali:
Committing changes related to VisaPermissions CRUD operation endpoints implementation.
#680


src/main/resources/flyway/sql/V1_23__add_visa_pcl.sql
```diff
@@ -1,8 +1,8 @@
...
-  entity                  UUID,
-  visaId                  UUID,
-  mask                    ACLMASK NOT NULL,
-  FOREIGN KEY (entity)    REFERENCES POLICY(id),
-  FOREIGN KEY (visaId)    REFERENCES GA4GHVISA(id)
+  policy_id               UUID,
+  visa_id                  UUID,
+  access_level            ACLMASK NOT NULL,
+  FOREIGN KEY (policy_id) REFERENCES POLICY(id),
+  FOREIGN KEY (visa_id)    REFERENCES GA4GHVISA(id)
```
joneubank, line 7: this naming is so much clearer. thanks for cleaning that up.

joneubank (approved):
Thanks! 👍

PR #696, opened 2023-05-09
Adding Client Id in the helm values as env variables

Azher2Ali:


joneubank:
We don't want to provide a client-id as a default. This is intentionally a placeholder value for other developers to modify.

Issue #708, opened 2023-05-29
BUG - EGO : All Admin access api's should not return HTTP 400 when invalid or not provided


Azher2Ali:
EGO : All Admin access api's should not return HTTP 400 when api key is not provided or invalid and should return HTTP 401

## Description
EGO : All Admin access api's should not return HTTP 400 when api key is not provided or invalid and should return HTTP 401. Also invalid jwt should return 401, valid jwt but wrong credentials (ie. not admin) should return 403

## Expected Behaviour
All Admin access api's in ego should return HTTP 401 Unauthorized when api key is not provided or is invalid.

## Actual Behaviour
All Admin access api's in ego is returning HTTP 400 Bad Request when an api key is not provided or is invalid.

Sample response
{
  "timestamp": "2023-05-26T14:32:48.744+0000",
  "status": 400,
  "error": "Bad Request",
  "path": "/visa"
}

## Possible Fix
Need to compare with older ego version to check the difference in changes done during Spring version upgrade for OAuth2SecurityRequestResolver

## Steps to Reproduce
1. Go to 'https://ego-passport.dev.cancogen.cancercollaboratory.org/swagger-ui/index.html'
2. Do not provide any api key in "Authorize" tab.
3. Test any api within ego
4. Response returned is HTTP 400 Bad Request

PR #710, opened 2023-06-01
Changes related to error status 400 and updates to deletion of visa

Azher2Ali:
As mentioned in the earlier ticker #708 ,The issues related to API's returning HTTP 400 error status has now been fixed. With the recent changes, if any user does not authenticated with valid jwt the error code returns back as HTTP 401 (Unauthorized) with a error status {
  "error": "Unauthorized",
  "message": "Required request header 'Authorization' for method parameter type String is not present",
  "timestamp": "2023-06-01T16:00:02.624+0000",
  "path": "/visa"
}

Steps to reproduce: 

1. Go to 'http://localhost:8081/swagger-ui/index.html#/Visa/deleteVisa'
2. Do not provide any api key in "Authorize" tab.
3. Test any api within ego
4. Response returned is HTTP 400 Bad Request

**Below is the ss, displaying the error status**
Screenshot 2023-06-01 at 1 20 09 PM

The second change is regarding the updates to deletion of visa. Earlier, there was an option to delete the Visa with using VisaId as a parameter. Now, as we are no longer using the VisaId to fetch or update. Therefore, "type" and "value" can be used to delete the Visa now.

### DIFF
--- a/src/main/java/bio/overture/ego/model/exceptions/ExceptionHandlers.java
+++ b/src/main/java/bio/overture/ego/model/exceptions/ExceptionHandlers.java
@@ -1,8 +1,7 @@
 package bio.overture.ego.model.exceptions;
 
 import static java.lang.String.format;
-import static org.springframework.http.HttpStatus.BAD_REQUEST;
-import static org.springframework.http.HttpStatus.NOT_FOUND;
+import static org.springframework.http.HttpStatus.*;
 
 import bio.overture.ego.utils.Joiners;
 import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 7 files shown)


src/main/java/bio/overture/ego/service/VisaService.java
```diff
@@ -81,9 +78,22 @@ public List<Visa> getByTypeAndValue(@NonNull String type, @NotNull String value)
     }
   }
 
-  public void delete(@NonNull UUID id) {
-    checkExistence(id);
-    super.delete(id);
+  public List<Visa> getByTypeAndValueForPassport(@NonNull String type, @NotNull String value) {
+    val result = visaRepository.getByTypeAndValue(type, value);
```
leoraba: this method logic looks duplicated.. is it possible to reuse it and call it from getByTypeAndValue method?

src/main/java/bio/overture/ego/model/exceptions/ExceptionHandlers.java
```diff
@@ -35,6 +35,21 @@ public ResponseEntity<Object> handleConstraintViolationException(
         NOT_FOUND);
   }
 
+  @ExceptionHandler(MissingRequestHeaderException.class)
```
leoraba, line 38: 👍

leoraba (approved):

PR #711, opened 2023-06-03
Update README.md

ciaranschutte:


MitchellShiell:
Good catch will check and update the rest of Readmes first thing Monday 
- Also noticed I need to update some text here

MitchellShiell (approved):

justincorrigible:
if it’s ok with y’all, i’d like to figure out what happened to the subdomain first… it may be an easier single fix than changing readmes everywhere. 

@henro001 any ideas? I don’t have visibility for the NS

justincorrigible:
Think I've found something to look into, when Collab is back up:
https://github.com/overture-stack/website/issues/210#issuecomment-1442291093
I have a hunch the docker container has simply been down since the last time system "reboot", and the solution will be to bring this `slackin` thing into the fold of Overture components we provision through Jenkins/Terraform.

